PR #44, opened 2015-02-12
ISO format milliseconds are preserved

alexkwolfe:
The millisecond component of an ISO formatted string is no longer discarded


wanasit:
I don't anyone who needs millisecond level accuracy, but thanks!

PR #51, opened 2015-04-18
Support for `a` and `an`

penne12:
Something I did through the web interface, so hopefully it works (I didn't have time to test it, more of a concept).

Should think of `a` and `an` as 1


wanasit:
Thanks for the improvement.

penne12:
Your welcome! I'm glad that worked.

PR #55, opened 2015-05-12
Check for 'a.m.'/'p.m.' does not work

marleyblakedog:
String.replace only replaces the first occurrence of a string, yielding 'am.'. Change to a regex so we can do a replace all.


wanasit:
Cool! Thanks for the fix.

Issue #58, opened 2015-06-11
Can't parse 05-25-2015


sean-hill:
Title says it all :smile:

sean-hill:
Thank you! :+1:

PR #61, opened 2015-08-09
Fixed https://github.com/wanasit/chrono/issues/58 dash was not allowed

wickedest:
Also changed to recognize unambiguous dd-mm-yyyy.
Also simplified regexp for day of week.


wanasit:
Thanks for the improvement! 
It would be nice if there are some tests for `until | through | till|` that you added, but I will find a chance to add that later.

PR #68, opened 2015-10-27
allow am or pm to be shorten to a or p

nevafuse:


nevafuse:
Not exactly sure which test failed, but it is passing for me locally.  I did add a test case to the same set that is failing.  Maybe it is running my test case against the old code?  If that is the case, that is why it is failing.

wanasit:
Thanks!

PR #81, opened 2016-01-30
Use keys from DAYS_OFFSET in the regular expression, so adding keys can

gregeinfrank:
be done in one place


penne12:
+1

wanasit:
I like the idea. Thanks.

Issue #43, opened 2015-02-02
Default to upcoming day of the week


j6k4m8:
There is perhaps a good reason for this behavior, so forgive me if this is intentional! My thought is that something like `Friday` should 'assume' `[this] Friday`, in that I am more likely to refer to an upcoming date than a past date by the day-of-the-week alone. (I think one would be more likely to refer to a past Friday as `last Friday`, even if it was only two or three days ago.) 

I'm looking through the code now to see if I can PR; let me know if I should abandon!

wanasit:
It really depends on the application and the type of text that you parse (Appointments, News, Emails, etc). 

So, I don't think it good idea to assume that everything is upcoming date. Instead, I currently compare the closest 7-days window (from the reference date) to calculate default weekday, which I **personally** think quite reasonable for most domain. 

However, I'm not going to stop you if you want to customize it. I would be happy to see your pull-request.

PR #82, opened 2016-02-01
Forward dates only when implying weekday

gregeinfrank:
In response to https://github.com/wanasit/chrono/issues/43 - this is a feature we also needed. Defaulting to the past doesn't work for scheduling, etc So I'm sharing this in hopes that others find it useful.

If you want to find the closest upcoming monday, but the closest monday was in the past the current implementation will choose the monday in the past.

This defaults everything to today or in the future. If you want the past using a weekday, you have to type "last monday", etc


penne12:
+1 - love this, and it'll defiantly help with my use case.

j6k4m8:
a *huge* +1 from me as well.

penne12:
@gregeinfrank Could you make this configurable, please?

gregeinfrank:
@penne12 I've updated to make `forwardDatesOnly` a config option cc/ @wanasit

penne12:
I've merged it - thanks$

sharpshark28:
Seems like a configurable option that should apply beyond just weekdays, such as "Dec 5" would select Dec 5 2015 because it's closer than Dec 5 2016.

PR #86, opened 2016-02-18
Added rule to parse 'tom' as 'tomorrow'

dreshfield:
Relatively simple change to add a more intuitive shortcut for 'tomorrow'.


penne12:
This seems like it would break people's usage:

Meeting with Tom at 4:00 PM
Buy Tom's Shoes Saturday

I'll see if I can get around to making the words used for parsing tomorrow configurable, though.

PR #93, opened 2016-02-19
Support "now" in spanish

gimenete:
I didn't include that in the previous PRs because it was related with two of them (supporting spanish and supporting "now"). Now that both are merged I suggest this additional change.

Thanks!


penne12:
Merged ! Thanks!

PR #107, opened 2016-03-26
Make now accurate to milliseconds

chuym:
`now` date parser is accurate to milliseconds, so it's a bit more `now` than it currently is.


penne12:
Thanks!

PR #108, opened 2016-04-01
Add parsing for seconds

mgduk:
This adds the ability to parse for seconds in English “time ago” and “deadline” formats.

It just follows the same pattern as minutes/hours; nothing novel.

Tests all still pass.


penne12:
Hey @mgduk! Mind writing tests for your change? Natural language is very finicky, and we don't want to accidently break your change in the future.

PR #136, opened 2016-09-02
added optional 'on' phrase to en weekday recognition

victor36max:
'on' is a common preposition before english weekday, i think we should parse it as well


wanasit:
Thanks

PR #139, opened 2016-09-09
Add example use of ParsedComponents functions to README

mvolz:


wanasit:
Thanks for adding this.

PR #142, opened 2016-09-22
Updated RegEx in SlashDateFormatParser for better handling

Schultzer:
This resolves #143.

Added test for European phone number and dash dates
atom removed whitespaces

@wanasit 👀 any comment on this?

PR #145, opened 2016-10-24
Added German Translation

matthias-christen:
Added German translation an tests (includes fix #142).

Note: Some tests in English containing the verb "do" had to be modified, because "Do" is also the abbreviation for "Donnerstag" (Thursday).


penne12:
Can you get someone who speaks German to verify that your changes make sense? If so, I'd be happy to merge.

wanasit:
Thanks for your great work. I want to play around a little bit before merging this.

First, I have the same feeling as Ben, we have tested it well enough, e.g. [from coverage report](https://coveralls.io/builds/8475221/source?filename=src%2Frefiners%2FDE%2FDEMergeDateTimeRefiner.js), we haven't tested merging date and time in German even once. 

Second, I worry about "do" as Donnerstag. After the change, every time we see word "do" we will guess it as Thursday. That's probably too much. I think we should allow user to select languages or don't use "do" abbv as casual date parsing.

penne12:
I think making languages selectable would be the best idea.

jpenna:
I'm doing the portuguese translation and I ran through this "language selection" problem, it was using the French refiner before it reached portuguese, so I changed orders (which is not the best approach).
So I'm thinking of doing this "language selection" feature, do you have anything done in this matter already?

wanasit:
@penne12 @jpenna Hi. Sorry for my very slow response. I was busy with my day job. I have an idea for a better languages support and will work on it in soon. 

Users should be able to customize chrono languages like this.
```javascript
var options = chrono.options.mergeOptions([
	chrono.options.en.casual(),
	chrono.options.en.strict(),
	chrono.options.de(),
	// ...
])

var customChrono = new chrono.Chrono(options);
customChrono.parse(...)
```

We should also make a language shortcut available.
```javascript
chrono.parse('Today at 1 AM')
chrono.de.parse('Freitag 30.12.16');
```

When we add new languages, the `options.js` file should look like this.
```javascript
exports.de = function () {
	
	parsers: [
		...
	],

	refiners: [
		...
	]
}

exports.de.strict = function () {
	
	parsers: [
		...
	],

	refiners: [
		...
	]
}
```

Any comment?

benaubin:
@wanasit Sounds good!

PR #147, opened 2016-10-24
Added capability to parse AD/BC in year (English, French, Spanish)

matthias-christen:
In the French and Spanish translations the "BE" (= Buddhist Era) (after the year number) in English was misinterpreted to mean "BC" (= Before Christ). Added and corrected support for "BC" / "AD" in English, French, Spanish.

In French, the abbreviations
- AC
- a. C. n., a. Chr. n.
- p. C. n., p. Chr. n.
are accepted. The fullstops are optional. The abbreviations are not case sensitive.

In Spanish, the abbreviations
- a. C.
- a.d.
- d. C.
are accepted. The fullstops are optional. The abbreviations are not case sensitive.


wanasit:
Merging this. Thanks for the improvement.

PS: I don't think someone would create a calendar appointment sometime in BC though :)

penne12:
@wanasit I have a differing opinion about people creating calendar appointments in BC. Want to talk about this January 9th, 100 BC?

matthias-christen:
Thanks for accepting the pull request 😄 
Indeed, no one might ever use this, but while doing the German translation (which we need for a project) I stumbled upon the "BE" in English, which was interpreted as "AC" (="BC") in French and Spanish, so I thought, why not do it right if we're half way there already 😉 
You never know, though, what people will use chrono for; I think it isn't limited to scheduling appointments (we use it to filter items by date) 😉 
Thanks everyone for the work on this great library!

penne12:
@matthias-christen Totally agree with you! Thanks you for the nice words.

PR #149, opened 2016-11-04
Support "next/last/past (X) week(s)" date formats

larissa:
Add support for expressions such as:

 - Next Week
 - Past 2 weeks
 - Next few days

Also support "past monday" as a synonym of "last monday"


larissa:
Existing tests had to be fixed because they have string that matches the new supported pattern and that's what was being matched first thus not giving the result the test expected. One way of the other the tests needed to be corrected and I didn't modify what was being tested since what I removed had no relation with what those test are supposed to test.

I defined a few and assumed it to be 3 because Chrono already does that with other patterns, so it's behavior consistency. I can remove it though, no problem.

penne12:
> Existing tests had to be fixed because they have string that matches the new supported pattern and that's what was being matched first thus not giving the result the test expected. One way of the other the tests needed to be corrected and I didn't modify what was being tested since what I removed had no relation with what those test are supposed to test.

Yeah, but changing features is a breaking change, and we can't ethically do that to add a feature (semantic versioning). Can you use a refiner to avoid parsing `last week` in the existing scenarios?

> I defined a few and assumed it to be 3 because Chrono already does that with other patterns, so it's behavior consistency. I can remove it though, no problem.

I would prefer if you make a `few` configurable instead of removing it outright (with a default set to 3). If you need help with doing so, I'd be willing to pitch in. I'm about to start working on making past options configurable, actually. 

Lastly, thanks so much for your PR

larissa:
>Can you use a refiner to avoid parsing last week in the existing types of scenarios (honestly not sure, if not, I'll see what I can do)?

I believe I can write one to prioritize the more specific date.

> I would prefer if you make a few configurable instead of removing it outright (with a default set to 3). If you need help with doing so, I'd be willing to pitch in. I'm about to start working on making past options configurable, actually.

Do you mean using an option `chrono.parse(text,new Date(),{setFew: 4});`
If not, I'm not sure how I would make it configurable so if you have a direction to point to me I could do that.

penne12:
> I believe I can write one to prioritize the more specific date.
Awesome

> Do you mean using an option chrono.parse(text,new Date(),{setFew: 4});
If not, I'm not sure how I would make it configurable so if you have a direction to point to me I could do that.

Yeah. I'll create an example to show you :)

larissa:
@penne12 Honestly I feel like there is no good way of making a refiner to accomplish what you want without:
1. changing _anything_ on old tests. If we're merging the two patterns and using the more specific date, at the very least the index of the matched result is going to change.
1. introducing a much more intrusive change. To avoid that I had to apply the refiner only if one of the matches have the tag for "next week" type of pattern, which is not using the full potential of the refiner but it's what I can do for now.

A _non good way_ would be just changing the order of the matches on the result array so that's the one that will be tested against the old test, but that goes against the idea of the results array containing the results in the order they were found. We can also just remove the result for the "next week" pattern, which I don't think it's a good solution either.

penne12:
The reference date is fine - it's used to make sure the test doesn't fail, even if ran years after it was written (why the test works normally).

Feel free to change the order of the results array, as long as you properly document the behavior in the README and you test it properly.

larissa:
Hi @penne12. You misunderstood me. I know why the reference date is there, what I meant is that in that context, in that text the date is incorrect (ie, doesn't make sense) given the reference date. The text say "next year (Jul-2017)" implying that next year __is__ 2017 (that's what the refinier is supposed to do, it combines the two dates if they refer to the same thing). But since the reference date is in 2012, "next year" in this case would be 2013 and not 2017, so the test needed to be corrected without changing its meaning.

I finished the refiner, it combines the two patterns if one is more specific than the other: 

* "She is getting married next year (July 2017)." - The second pattern is more specific because it has a month, while the first only has a year, so the refiner would merge "next year" with "July 2017" and use the `result.start` from "July 2017"
* "The event is going ahead next month (04/2016)" - Neither is more specific (both have certain month and year, but implied day), so it doesn't combine anything and returns two results.

This is a better approach than simply changing the order.

larissa:
hi @penne12. Did you have opportunity to take a look at this PR after I included the refiner?

src/options.js
```diff
@@ -9,6 +9,7 @@ function baseOption(strictMode) {
             // EN
             new parser.ENISOFormatParser(strictMode),
             new parser.ENDeadlineFormatParser(strictMode),
+            new parser.ENRelativeDateFormatParser(strictMode),
```
wanasit: Relative date format is casual by nature, and should not by be in strict mode in the first place. Instead of supporting strict mode, just don't include the parser into strict configuration is probably better.

src/parsers/EN/ENRelativeDateFormatParser.js
```diff
@@ -0,0 +1,118 @@
...
+    Parser.apply(this, arguments);
+
+    this.pattern = function() {
+        return this.isStrictMode()? STRICT_PATTERN : PATTERN;
+    };
+
+    this.extract = function(text, ref, match, opt){
+
+        var index = match.index + match[1].length;
+        var modifier = match[2].toLowerCase().match(/^next/) ? 1 : -1;
```
wanasit: Instead of using group number directly, could you define the constant first? It's easier to debug.
```
var MODIFIER_WORD_GROUP = 2;
var MULTIPLYER_WORD_GROUP = 3;
...
```

src/parsers/EN/ENWeekdayParser.js
```diff
@@ -12,30 +12,30 @@ var DAYS_OFFSET = { 'sunday': 0, 'sun': 0, 'monday': 1, 'mon': 1,'tuesday': 2, '
...
+    '(?:\\s*(this|last|past|next)\\s*week)?' +
     '(?=\\W|$)', 'i');
 
 var PREFIX_GROUP = 2;
 var WEEKDAY_GROUP = 3;
 var POSTFIX_GROUP = 4;
 
 
-exports.updateParsedComponent = function updateParsedComponent(result, ref, offset, modifyer) {
+exports.updateParsedComponent = function updateParsedComponent(result, ref, offset, modifier) {
```
wanasit: thanks ^^

src/refiners/EN/ENPrioritizeSpecificDateRefiner.js
```diff
@@ -0,0 +1,109 @@
...
+var ParsedComponents = require('../../result').ParsedComponents;
+var Refiner = require('../refiner').Refiner;
+
+
+var PATTERN = new RegExp("^\\s*(at|after|before|on|,|-|\\(|\\))?\\s*$");
+
+function isMoreSpecific(prevResult, currResult) {
+    var moreSpecific = false;
+
+    if (prevResult.start.isCertain('year'))
```
wanasit: Could you please add `{ }` on all `if` and `else` clue, even those one-line ones.

test/test_en_middle_endian.js
```diff
@@ -1,21 +1,21 @@
 
 test("Test - Single Expression", function() {
-    var text = 'She is getting married next year (July 2017).';
+    var text = 'She is getting married next year (July 2013).';
```
wanasit: Instead of updating the test here, I prefer you remove "next year". And add a new test case for "next year (July 2013)" in test_en_relative.js

wanasit: similar to the one you do in test_en_slash.js

src/refiners/EN/ENPrioritizeSpecificDateRefiner.js
```diff
@@ -0,0 +1,109 @@
...
+        if (!currResult.start.isCertain('year'))
+            moreSpecific = true;
```
wanasit: Could we ever hit this case? You always assign a year right?

test/test_en_relative.js
```diff
@@ -0,0 +1,182 @@
...
+
+    var result = results[0];
+    if(result){
+        ok(result.text == text, result.text);
+        ok(result.start.get('year') == 2017, JSON.stringify(result.start));
+        ok(result.start.get('month') == 2, JSON.stringify(result.start));
+        ok(result.start.get('day') == 1, JSON.stringify(result.start));
+        ok(result.start.get('hour') == 12, JSON.stringify(result.start));
+    }
+
```
wanasit: Could you please add another test "next week (Dec 2016)"?

src/refiners/EN/ENPrioritizeSpecificDateRefiner.js
larissa: Actually for "next week" I don't assign a year, since a change of year/month can occur in the next week. For example if today was 25 Dec 2017, next week could be 31 Dec 2017 or 1 Jan 2018.

larissa:
@wanasit Thanks for the feedback! I'm applying the changes you asked now.

PR #150, opened 2016-11-09
list CDNJS in the readme

kennynaoh:
This can help user quickly access script tag from cdnjs


penne12:
Don't recommend using cdn.js - just host the file yourself (stabler website, etc)

PR #151, opened 2016-11-13
added fr culture's first day of month parsing

krazylek:
Quite often the 1st day of month is written "1er" (following the language)

I added a test for the case, but sorry forgot to create a branch!

Cheers

### DIFF
--- a/src/parsers/FR/FRMonthNameLittleEndianParser.js
+++ b/src/parsers/FR/FRMonthNameLittleEndianParser.js
@@ -14,7 +14,7 @@ var DAYS_OFFSET = util.WEEKDAY_OFFSET;
 
 var PATTERN = new RegExp('(\\W|^)' +
         '(?:(Dimanche|Lundi|Mardi|mercredi|Jeudi|Vendredi|Samedi|Dim|Lun|Mar|Mer|Jeu|Ven|Sam)\\s*,?\\s*)?' +
-        '([0-9]{1,2})' +
+        '([0-9]{1,2})(?:er)?' +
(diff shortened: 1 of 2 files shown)


test/test_fr_little_endian.js
```diff
@@ -22,6 +22,26 @@ test("Test - Single expression", function() {
     }
 
 
+    var text = "1er Août 2012";
```
Schultzer: there shuld be test for edge case like `20er Août 2012` and `0er Août 2012` ect.

src/parsers/FR/FRMonthNameLittleEndianParser.js
```diff
@@ -14,7 +14,7 @@ var DAYS_OFFSET = util.WEEKDAY_OFFSET;
 
 var PATTERN = new RegExp('(\\W|^)' +
         '(?:(Dimanche|Lundi|Mardi|mercredi|Jeudi|Vendredi|Samedi|Dim|Lun|Mar|Mer|Jeu|Ven|Sam)\\s*,?\\s*)?' +
-        '([0-9]{1,2})' +
+        '([0-9]{1,2})(?:er)?' +
```
Schultzer: I don't like this what about `20er` and `0er` etc.

krazylek: Hi, yes I agree about the other days

I mainly followed what was done for other cultures ("2st" is legal value), but right, here in french it's only valid for the first day of month: "2er" would be wrong, but "21er" as well.

So it could probably changed to:

```
([0-9]{1,2}|1er)
```

Do you think the change is alright? 

And probably wouldn't require a refiner in this case

penne12:
Looks great. Thanks for the PR @krazylek

PR #153, opened 2016-11-20
Separate "morning" "afternoon" etc logic

guillegette:
I found that with the current set up we were able to only do things like "tomorrow morning" or "yesterday afterrnoon" but things like "monday morning" or specific dates with "morning" would not work. 

My suggested solution was to create a casual time parser only for those terms like "morning", "afternoon", etc. We can add more later like "lunch time" "sunset" etc etc

Test passed 100% and I created a test for the day of the week and casual time.

Let me know your thoughts.

### DIFF
--- a/src/options.js
+++ b/src/options.js
@@ -78,6 +78,7 @@ exports.casualOption = function () {
 
     // EN
     options.parsers.unshift(new parser.ENCasualDateParser());
+    options.parsers.unshift(new parser.ENCasualTimeParser());
     options.parsers.unshift(new parser.ENWeekdayParser());
     options.parsers.unshift(new parser.ENRelativeDateFormatParser());
 
@@ -91,6 +92,6 @@ exports.casualOption = function () {
     // FR
     options.parsers.unshift(new parser.FRCasualDateParser());
     options.parsers.unshift(new parser.FRWeekdayParser());
-    
+
     return options;
 };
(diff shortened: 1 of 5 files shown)


guillegette:
Update: using the opt argument you can define what "morning", "afternoon", etc means for you within your application logic. 

Example: `chrono.parse("Lets meet on Tuesday morning", new Date(2012, 7, 10, 12), {"morning":8})`

```
ParsedResult {
  ref: 2012-08-10T19:00:00.000Z,
  index: 10,
  text: 'on Tuesday morning',
  tags:
   { ENWeekdayParser: true,
     ENCasualTimeParser: true,
     ENMergeDateAndTimeRefiner: true },
  start:
   ParsedComponents {
     knownValues: { weekday: 2, hour: 8, minute: 0, second: 0 },
     impliedValues: { day: 7, month: 8, year: 2012, millisecond: 0 } } }
```

penne12:
Looks great! Thanks for the PR.

guillegette:
Awesome! Thanks for merging.

Slapbox:
@guillegette @benaubin - I know this is old an old PR, but I assume this still be working? No matter what I pass I get 6am for morning. I don't see much in the way of discussion of valid parser options besides `timezone` and `forwardDate`

Slapbox:
@wanasit should customizing `morning`, `afternoon`, etc, still work? We can't get it to work, but it seems to be present still? Any advice would be appreciate greatly.

PR #158, opened 2016-12-13
Add CDNJS version badge in readme

kennynaoh:
This will add the badge to show its version on CDNJS and also link to its page on CDNJS!


penne12:
Much rather have a npm version badge - it's more accurate, and you should almost always self host your external sources. NPM allows more flexibility and doesn't make us depend on a external source.

PR #164, opened 2017-01-06
Fix order of dashed date for spanish

francof:
The dates in spanish when using dashed format usually go dd/mm/yyyy, not mm/dd/yyyy.

This PR fixes the parser and the corresponding tests.


gimenete:
LGTM. Thanks!

penne12:
@Schultzer I agree - we can't change a test without breaking compatibility, and I don't think this change is worth the tradeoff.

francof:
Sorry guys, was AFK, I will revert soon

penne12:
@francof Don't worry about it. Thanks so much for contributing!

PR #184, opened 2017-04-05
fix markdown errors

wellsjo:
using javascript syntax highlighting doesn't work for the repl


benaubin:
actually, syntax highlighting as-is works fine. I don't see how changing from JS to bash would help.

PR #188, opened 2017-04-08
fix: French months can have accents, but also accept the non-accentua…

denouche:
…ted versions


denouche:
Added tests and rebased the branch.
Thanks!

wanasit:
Merged. Thanks!

PR #189, opened 2017-04-09
feat: add FRRelativeDateFormatParser

denouche:


denouche:
added test, and rebased the branch.

wanasit:
Thank you! @denouche
 
This looks good to me. Would be nice if you could add more test for second, minute, quarter and year to make the coverage go higher. But I think this is ok already.
https://coveralls.io/builds/11491008/source?filename=src%2Fparsers%2FFR%2FFRRelativeDateFormatParser.js

@benaubin do you have more comment?

denouche:
Found a bug while adding one more test. I will update the tests.

denouche:
I added the tests for second, minute, quarter and year. 

By the way thanks for this project :+1:

denouche:
wow, the error is:

```> qunit -l '{summary:true, tests: true}' -c chrono:src/chrono.js -p src -t test/*.js --cov
[Error: Process blocked for too long]```

Any idea?

denouche:
I propose to abandon  your node-qunit fork and to take the 0.9.3 of the original repo, since they integrated your changes. Moreover they added a `--timeout` option to avoid the case we have now.

I make the changes on  this branch, in another commit.

wanasit:
Oh. Thanks for patching the qunit too. Merging this.

PR #207, opened 2017-07-19
Indicate the implied year when parsing e.g. '5/31'

AndrewKvalheim:
It surprised me that the year [chosen] for e.g. `5/31` is reported to be known rather than implied:

```javascript
const ref = new Date(1999, 6-1, 1);

// Expected
chrono.parse('May 31', ref)[0].start.isCertain('year') // => false
chrono.parse('5/31',   ref)[0].start.isCertain('year') // => false
chrono.parse('May 31', ref)[0].start.get('year') // => 1999
chrono.parse('5/31',   ref)[0].start.get('year') // => 1999
chrono.parse('May 31', ref, {forwardDatesOnly: true})[0].start.get('year') // => 2000
chrono.parse('5/31',   ref, {forwardDatesOnly: true})[0].start.get('year') // => 2000

// Actual
chrono.parse('May 31', ref)[0].start.isCertain('year') // => false
chrono.parse('5/31',   ref)[0].start.isCertain('year') // => true
chrono.parse('May 31', ref)[0].start.get('year') // => 1999
chrono.parse('5/31',   ref)[0].start.get('year') // => 1999
chrono.parse('May 31', ref, {forwardDatesOnly: true})[0].start.get('year') // => 2000
chrono.parse('5/31',   ref, {forwardDatesOnly: true})[0].start.get('year') // => 1999
```

Would it make sense to `imply` it instead so that `forwardDatesOnly` can be used here?

[chosen]: https://github.com/wanasit/chrono/blob/v1.3.1/src/parsers/EN/ENSlashDateFormatParser.js#L68

### DIFF
--- a/src/parsers/EN/ENSlashDateFormatParser.js
+++ b/src/parsers/EN/ENSlashDateFormatParser.js
@@ -116,7 +116,11 @@ exports.Parser = function ENSlashDateFormatParser(config) {
 
         result.start.assign('day', day);
         result.start.assign('month', month);
-        result.start.assign('year', year);
+        if (match[YEAR_GROUP]) {
+            result.start.assign('year', year);
+        } else {
+            result.start.imply('year', year);
+        }
 
         //Day of week
         if(match[WEEKDAY_GROUP]) {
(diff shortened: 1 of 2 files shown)


wanasit:
That's my mistake. Thank you for the fix.

PR #208, opened 2017-07-25
ENDeadlineFormatParser should have implied rather than known Date values

AlexDobeck:
Currently when a date in Deadline format is parsed it results in a series of known values:
``` 
{  
   "ref":"2017-07-25T23:11:17.234Z",
   "index":246,
   "text":"within 30 days",
   "tags":{  

   },
   "start":{  
      "knownValues":{  
         "year":2017,
         "month":8,
         "day":24
      },
      "impliedValues":{  
         "hour":12,
         "minute":0,
         "second":0,
         "millisecond":0
      }
   }
}
```

Since all the values from this parser are based on the Reference Date, I think it's important to indicate that these are all implied, rather than known.

As far as a use-case goes; I'm currently leveraging this project with historical documents where a reference date can not always be determined. To ensure I don't provide incorrect data, any date which has an implied year/month are removed. Since 'Within 30 days' is being parsed as Today + 30 days, but not indicating that the values are implied, I'm left manually removing these instances.

The changes consist of altering all 'assign' methods to 'imply' from the Deadline parser, and a set of tests around it. Only done to EN, I can apply similar changes to DE/ES/FR/ZHHant if desired.

### DIFF
--- a/src/parsers/EN/ENDeadlineFormatParser.js
+++ b/src/parsers/EN/ENDeadlineFormatParser.js
@@ -67,9 +67,9 @@ exports.Parser = function ENDeadlineFormatParser(){
                 date.add(num, 'year');
             }
 
-            result.start.assign('year', date.year());
-            result.start.assign('month', date.month() + 1);
-            result.start.assign('day', date.date());
+            result.start.imply('year', date.year());
+            result.start.imply('month', date.month() + 1);
+            result.start.imply('day', date.date());
             return result;
         }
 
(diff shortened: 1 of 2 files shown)


wanasit:
Thank you for the contribution.

And I agree that the date of future deadline e.g. "within 12 days" or "in 2 weeks" is not really determined and may should be treated as an "implied" rather than a "known". **I'm accepting this PR.**

However, I am not sure what is the benefit for correcting "imply" vs "assign" in this case. If you don't have correct ref date, you should just avoid using those parsers that rely on the reference altogether ("Monday", "last week", "3 days ago", etc). 

In Chrono, the implied value is only used internally as a way to say _the value is flexible to change depending on the context or what followed_. For example, Monday's date should be implied. The date can change depending on what follow ("Mon - Fri" vs "Sat-Mon"). I don't see there can be any additional information that could change the determination of the deadline patterns.

AlexDobeck:
You're probably right that it would be better to use the parsers that fit, and the Deadline Parser was definitely not written in a way to handle non-current documents.

Though, in my case, while the documents may be in the past (or the timeline may be unable to be determined), we expect to see plenty of recent documents as well.

I'd like to use Chrono to parse a document once, then use Chrono's output at a later point. If Chrono identifies my origination date (and it's recent), the output from the Deadline Parser could be very helpful. If not, I'm in the situation where I need to cull the implied output.

PR #241, opened 2017-12-28
Fixing an inconsistency in the dates being parsed

flavianh:
"8 Février" and "08/02" are exactly the same in French.

## Previous behavior

```javascript
chrono.parse("8 Février", new Date(2012,7,10));  // 2013/02/08, the correct date
chrono.parse("08/02", new Date(2012,7,10));  // 2012/02/08
```

## Behavior with this PR

```javascript
chrono.parse("08/02", new Date(2012,7,10)) === chrono.parse("8 Février", new Date(2012,7,10));  // 2013/02/08, the correct date
```


flavianh:
@wanasit do you know how long it will be before this makes it to the next release?

PR #244, opened 2018-02-20
Fixed missing var declaration which was throwing.

DanielBoa:
For some reason this was only throwing when my project was being bundled with Webpack, not sure if that's relevant or not.  Maybe because "use strict" is added to the bundle?

Regardless it appears to be a bug as index isn't explicitly declared.


wanasit:
Ops. Thank you for the fix.

DanielBoa:
@wanasit When you get a chance could you publish this change to npm?

PR #248, opened 2018-03-15
Update moment and add nsp security check to tests

mvolz:
The currently pinned version of moment had several
security issues. This change updates moment to the
most recent version. It also adds nsp to npm test
to check for security vulnerabilities.


wanasit:
@mvolz Thank you for the fix and apologize that I didn't spend time maintaining the project as much as I should.

Issue #250, opened 2018-04-02
Casual Time should also imply AM/PM time


ubarsaiyan:
If you put "Let's meet tomorrow at 5", that goes to 5 AM, which is OK. 
If you put "Let's meet tomorrow evening", that goes to 6 PM, which is again OK. 
If you combine those two and put "Let's meet tomorrow evening at 5", it'll override the "evening" part and just consider the "at 5" part, which is not OK.

PR #251, opened 2018-04-02
Added correct parsing of phrases such as '2 in the evening'

ubarsaiyan:
Fixes issue #250 
Added correct parsing of phrases such as '2 in the evening'
Let's meet at 8 in the evening
What about 3 in the afternoon?
Let's meet tomorrow at 5 in the evening
Let's meet at 11 in night
Let's meet in the night at 11

All the tests pass too!


wanasit:
Thank you for your suggestion. I fixed it a bit different in 28737b3.
* You should add the test cases for the text you mentioned
* imply() instead of assign() would be better as it allows other part to override this e.g. "tonight at 1am".

Issue #259, opened 2018-05-09
day after tomorrow bug


myriem2424:
there is a bug with the " day after tomorrow ", chrono cannot deduct the correct date and only give the tomorrow date.

willlma:
I just came across a similar bug. The same is true of _the day before yesterday_.

willlma:
This is a duplicate of #170

tigercosmos:
the Chinese word "後天", which is the same as "day after tomorrow", is also not valid for parsing.

PR #260, opened 2018-05-17
ZH: "the (two) day before yesterday" & "the (two) day after tomorrow"

tigercosmos:
Support Chinese Word for
- "the day before yesterday", which is "前天|前日"
- "the two day before yesterday", which is "大前天|大前日"
- "the day after tomorrow", which is "後天|後日"
- "the two day after tomorrow", which is "大後天|大後日"

I add the tests for this change.
I have run test, and pass all.

also build `npm run make`

this is part of #259, but in Chinese


wanasit:
Thanks for the improvement.

PR #277, opened 2018-11-29
Add English support for relative phrases like "in 1.5 hours".

hryanjones:
First of all, thanks for creating this project!

I use it in a [small web reminder utility](https://hryanjones.com/webminder/) that I've been working on.

I found that I was very commonly writing strings like "in 1.5 hours" and a bit frustrated that they didn't work, so I've added support for this. I'll try to add comments to the PR to explain some of the reasons for things that may not be clear.

Thanks again!

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,6 +6,10 @@
     "type": "git",
     "url": "https://github.com/berryboy/chrono.git"
   },
+  "jest": {
+    "verbose": true,
+    "testURL": "http://localhost/"
+  },
   "license": "MIT",
   "version": "1.3.10",
   "directories": {
--- a/src/parsers/en/ENTimeLaterFormatParser.js
+++ b/src/parsers/en/ENTimeLaterFormatParser.js
@@ -4,14 +4,20 @@ var ParsedResult = require('../../result').ParsedResult;
 var util  = require('../../utils/EN');
 
 var PATTERN = new RegExp('' +
-    '(\\W|^)' +
-    '(' + util.TIME_UNIT_PATTERN + ')' +
-    '(?:later|after|from now|henceforth|forward|out)(?=(?:\\W|$))', 'i');
+    /*match[1]*/ '(\\W|^)' +
+    /*match[2]*/ '(in )?' +
+    /*match[3]*/ '(' + util.TIME_UNIT_PATTERN + ')' +
+    /*match[4]*/ '(later|after|from now|henceforth|forward|out)?' +
+    /*match[5]*/ '(?=(?:\\W|$))',
+'i');
 
 var STRICT_PATTERN = new RegExp('' +
-    '(\\W|^)' +
-    '(' + util.TIME_UNIT_STRICT_PATTERN + ')' +
-    '(?:later|from now)(?=(?:\\W|$))', 'i');
+    /*match[1]*/ '(\\W|^)' +
+    /*match[2]*/ '(in )?' +
+    /*match[3]*/ '(' + util.TIME_UNIT_STRICT_PATTERN + ')' +
+    /*match[4]*/ '(later|from now)?' +
+    /*match[5]*/ '(?=(?:\\W|$))',
+'i');
 
 exports.Parser = function ENTimeLaterFormatParser(){
     Parser.apply(this, arguments);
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -6,6 +6,10 @@
     "type": "git",
     "url": "https://github.com/berryboy/chrono.git"
   },
+  "jest": {
```
hryanjones: This was to fix [this error](https://stackoverflow.com/questions/51554366/jest-securityerror-localstorage-is-not-available-for-opaque-origins#answer-51554619) which was causing all the unit tests to fail.

I figure it'd be good to add to the project to make it a bit easier in case others run into this.

src/parsers/en/ENTimeLaterFormatParser.js
```diff
@@ -21,20 +27,25 @@ exports.Parser = function ENTimeLaterFormatParser(){
     };
 
     this.extract = function(text, ref, match, opt){
-
         if (match.index > 0 && text[match.index-1].match(/\w/)) return null;
 
+        var prefix = match[2];
+        var suffix = match[4];
+        if (!prefix && !suffix) return null;
```
hryanjones: it didn't seem worth worrying about when both the prefix and suffix were there, it makes sense in some phrases, but not all:
-  (a little weird, but seems fine):
  - "in 3 minutes later"
  - "in 3 minutes henceforth
- (seems good) "in 3 minutes from now"

etc.

src/utils/EN.js
```diff
@@ -111,12 +111,12 @@ exports.ORDINAL_WORDS_PATTERN = '(?:'
     + ')';
 
 var TIME_UNIT = 
-    '(' + exports.INTEGER_WORDS_PATTERN + '|[0-9]+|an?(?:\\s*few)?|half(?:\\s*an?)?)\\s*' +
+    '(' + exports.INTEGER_WORDS_PATTERN + '|[0-9]+|[0-9]+\.[0-9]+|an?(?:\\s*few)?|half(?:\\s*an?)?)\\s*' +
```
hryanjones: This and the `parseFloat` were what was needed to support fractional time chunks

src/utils/EN.js
```diff
@@ -111,12 +111,12 @@ exports.ORDINAL_WORDS_PATTERN = '(?:'
...
     '(sec(?:onds?)?|min(?:ute)?s?|hours?|weeks?|days?|months?|years?)\\s*';
 
 var TIME_UNIT_STRICT = 
-    '([0-9]+|an?)\\s*' +
-    '(seconds?|minutes?|hours?|days?)\\s*';
+    '(?:[0-9]+|an?)\\s*' +
```
hryanjones: I made these non-capturing groups so that both the strict and non-strict takes up the same number of sections within a match. Otherwise it'd need different indexes depending on strict mode or not and that seemed messier than needed.

test/en/en_time_later.test.js
```diff
@@ -69,12 +69,27 @@ test("Test - Single Expression", function() {
 });
 
 test("Test - Single Expression (Strict)", function() {
-
     var text = "15 minute after";
     var results = chrono.strict.parse(text, new Date(2012,7,10,12,14));
-    expect(results.length).toBe(0);
+    expect(results[0]).toBe(undefined);
```
hryanjones: I found this to be more useful than the existing pattern because it'd then pretty print the first result to help in trying to figure out which parser was hit and such.

It'd probably be good to do this everywhere in the tests. I can make that change if you agree.

test/en/en_time_later.test.js
```diff
@@ -69,12 +69,27 @@ test("Test - Single Expression", function() {
...
+        expect(result.index).toBe(0);
+        expect(result.text).toBe('in 25 minutes');
+        expect(result.start.get('hour')).toBe(13);
+        expect(result.start.get('minute')).toBe(5);
+
+        var resultDate = result.start.date();
+        var expectDate = new Date(2012, 7, 10, 13, 5);
+        expect(expectDate.getTime()).toBe(resultDate.getTime())
+    }
 });
```
hryanjones: Any other tests you think I should add?

gagan-bansal:
Great, each code change has been explained very well.

I am just user of 'chrono'. 
I have a question if this is still in development?
Chrono is very useful lib. So maintaining this will help a big  community.

hryanjones:
I'm not sure if it's still in development considering last commit was 6 months ago. Depending on the reasons for that, I'm hoping my pull request might be easy enough to just approve by the owner.

wanasit:
@hryanjones I apologize for my unresponsiveness and thank you very much for your change. Also, thank you for including the explanation and tests.

hryanjones:
You're welcome!

Please, do consider that change to the tests that allows the result to be pretty printed if it fails.

PR #278, opened 2018-12-05
Fix bug in ParsedResult#clone

brandonhorst:
ParsedResult.prototype.clone would create a clone but not return it.


wanasit:
Thank you for showing me. I fixed it differently in 21cba1b.

PR #279, opened 2018-12-16
Added PT locale

samin:
Added PT language to parser, based off of ES parser (languages are the most similar).


wanasit:
Thank you for the pull request and sorry for my slow reply. 

Could you please rebase the current master and run the unittest `npm test`?

PR #282, opened 2019-02-21
Update FRCasualDateParser.js

sullix:
* For "maintenant" (now), assign the time (hour, minute, second, millisecond) rather than just implying it.
* Another accepted writing for "après-midi" (afternoon) without dash.


benaubin:
@sullix Out of curiosity, why did you change from `assign` to `imply`?

wzhouwzhou:
Any news on this merge conflict?

benaubin:
This is stale and won't fit with v2, so I'm closing for now. Would be happy to re-review if made compatible with the current version of chrono.

PR #288, opened 2019-04-04
Added PT locale

samin:
Based off of ES locale. Some tests in english and french were failing, had to specify `chrono.en.parse` and `chrono.fr.parse` in some cases... don't know what the protocol is in these cases.


wanasit:
Thank you very much for the improvement.

PR #289, opened 2019-04-09
Loosen moment dependency

blowery:
Loosen the moment dependency so that downstream libraries can choose which version of moment to use.

Currently, consumers are forced to use `moment@2.21.0` if they use `chrono-node`, even if they're already using a more recent version. This causes moment, which uses a stateful singleton for things like the default locale, to load and initialize more than once.

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,6 +32,6 @@
     "coveralls": "npm run test && cat coverage/lcov.info | coveralls"
   },
   "dependencies": {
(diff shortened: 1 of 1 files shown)


wanasit:
Thanks for the improvement!

blowery:
@wanasit Thanks for landing this! Could you please release a new version to npm so we can take advantage of the change? Thanks!

PR #297, opened 2019-07-27
Implement 令和('reiwa' Japanese era) (2019〜) and 元年(the 1st year of the year)

gecko655:
440px-Yoshihide_Suga_announcing_new_imperial_era_Reiwa_2_(cropped)

Japanese emperor Akihito has retired on April 30th, 2019, and 令和 (reiwa) era has stared on May 1st, 2019.
In Japan, the 1st year of Reiwa is called "令和元年"(Reiwa-gannen) and never called "令和1年".
So now (2019-07-27) is called "令和元年7月27日"
https://en.wikipedia.org/wiki/Reiwa

I'd like to make the new era name "令和" and the form of era's 1st year "元年" parseable with chrono.

### DIFF
--- a/src/parsers/ja/JPStandardParser.js
+++ b/src/parsers/ja/JPStandardParser.js
@@ -9,7 +9,7 @@ var Parser = require('../parser').Parser;
 var ParsedResult = require('../../result').ParsedResult;
 
 var util  = require('../../utils/JP'); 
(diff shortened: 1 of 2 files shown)


wanasit:
ありがとございます。やまもりさん。

PR #298, opened 2019-07-27
 Implemented '本年' '今年' (this year) form for Japanese.

gecko655:
'本年7月27日' and '今年7月27日' form is used to express 'July 27th of this year' in Japanese.
This usage is exactly the same as the usage of '同年', which is already implemented in chrono.
(These 3 expressions differ in politeness and other kinds of contexts.

ref(本年): https://www.kantei.go.jp/jp/headline/gotaii_gosokui/sikiten.html
ref(本年): https://www.mhlw.go.jp/content/12400000/000475441.pdf
ref(今年): https://www.sankei.com/west/news/190323/wst1903230023-n1.html


gecko655:
I know this PR will conflict with #297.
I will fix the conflict after either of them merged.

wanasit:
Thanks for the improvement!

gecko655:
🙏

PR #300, opened 2019-08-10
Timezone aware moments

ParthKolekar:
Stealing commits from @xiangwgo and adding timezone aware moments into upstream, if you'd oblige to merge these.

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,6 +32,7 @@
     "coveralls": "npm run test && cat coverage/lcov.info | coveralls"
   },
   "dependencies": {
-    "moment": "2.21.0"
+    "moment": "2.21.0",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -32,6 +32,7 @@
     "coveralls": "npm run test && cat coverage/lcov.info | coveralls"
   },
   "dependencies": {
-    "moment": "2.21.0"
+    "moment": "2.21.0",
+    "moment-timezone": "0.5.17"
```
blowery: This is going to increase the download size for anyone using this package pretty dramatically. `moment-timezone` is not small.

wanasit:
I've replaced the Moment with DayJS from 1.4+. So, this change is not valid anymore.

PR #310, opened 2019-12-17
JpParser: Fixed parsing Feb 29th  leap year while current year is not a leap year.

gecko655:
When parsing '2020年2月29日' at this time (2019-12-17), chrono returns '2020-02-28'.
This is due to the order of registering 'year, month, day' to the `moment` object.

Month & day parts should be registered AFTER year part.
If month & day parts are registered BEFORE year part, 

- chrono tries to make '2019-02-29' `moment` object in the middle of making '2020-02-29'
- `moment` fall back to '2019-02-28' since '2019-02-29' is an invalid date
- then, chrono registers year part and returns '2020-02-28'

I also did some refactorings, that involve this issue.

### DIFF
--- a/src/parsers/ja/JPStandardParser.js
+++ b/src/parsers/ja/JPStandardParser.js
@@ -31,19 +31,6 @@ exports.Parser = function JPStandardParser(){
             ref: ref,
         });
         
-        var month = match[MONTH_GROUP];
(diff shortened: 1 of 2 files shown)


wanasit:
Thank you for reporting and fixing this. :)

PR #315, opened 2020-01-13
Update package.json to reflect GitHub repo transfer

domdomegg:
`berryboy` -> `wanasit`

Will make the npm page display this repo. Yes, it redirects but it's nicer to just be right in the first place, also allows searching NPM by repo as there are a few similarly named libraries on there.


wanasit:
Thanks.

Issue #339, opened 2020-07-27
 ParsingComponents.get typing wrong in dist/results.d.ts


Mattis6666:
When simply installing and running, the following error occurs
```ts
node_modules/chrono-node/dist/results.d.ts:9:5 - error TS2416: Property 'get' in type 'ParsingComponents' is not assignable to the same property in base type 'ParsedComponents'.
  Type '(component: Component) => number | undefined' is not assignable to type '(c: Component) => number | null'.
    Type 'number | undefined' is not assignable to type 'number | null'.
      Type 'undefined' is not assignable to type 'number | null'.

9     get(component: Component): number | undefined;
```

This is fixed by changing `node_modules/chrono-node/src/results.ts:28` from
```ts
    get(component: Component): number | undefined {
        if (component in this.knownValues) {
            return this.knownValues[component];
        }

        if (component in this.impliedValues) {
            return this.impliedValues[component];
        }

        return undefined;
    }
```
to 
```ts
    get(component: Component): number | null {
        if (component in this.knownValues) {
            return this.knownValues[component];
        }

        if (component in this.impliedValues) {
            return this.impliedValues[component];
        }

        return null;
    }
```
but I am not sure whether this would break anything else relying on that null.

PR #340, opened 2020-07-27
typing-fix: ParsingComponents.get: number | undefined -> number | null

Mattis6666:
Fixes wanasit/chrono#339


wanasit:
ok. Thanks for the fix.

PR #351, opened 2020-10-06
Adding NL locale

Pruxis:


wanasit:
@Pruxis Very cool! Thanks for working on this. 

It'd be nice if you could also add the tests for NL locale.
I could work on that later, but if you speak the language it would be nice if you could look into the tests.

PR #356, opened 2020-11-10
Fix for XX-YY dates not being parsed as dates

nkochakian:
Currently, dates in the MM-DD format are falling through the `SlashDateFormatParser` and are being incorrectly parsed as time strings by one of the time parsers later in the chain.


wanasit:
Hello. Thank you for the PR, but I’d have to refuse to merge this.

Treating dash as slash would result in too aggressive parsing for most use-cases. I don’t think, by default, `2-1 - 2-2` should mean `February 1st to February 2nd`.

PR #358, opened 2020-11-11
Fix time unit within using for

trothe:
Support texts like "set a timer for 5 minutes" or "wait for 1 hour".


wanasit:
Very clean patch + testing.
Thank you!

PR #359, opened 2020-11-11
Fix parsing of "May" using ENMonthNameParser

honzatrtik:
Matched string of length 3 and less was discarded, so that string `"May"` was not parsed correctly.


wanasit:
Thank you for you MR, but I decided to fix the problem differently in 8c0c24efa5113326730cebd4538793e06cd74eb9.

Your fix [failed unit tests](https://travis-ci.org/github/wanasit/chrono/jobs/743749565) on following cases
* `By Angie Mar November 2019`, expect `November 2019`, not `Mar`
* `jan`, expect empty

We need to separate full-month names (which are more likely) from the abbreviations (which are less likely) .

PR #361, opened 2020-11-14
Add ESLint

lostfictions:
This PR adds ESLint to this repo. ESLint catches a number of semantic errors that TypeScript does not (including some errors in TypeScript types themselves!) and it's a recommended best practice (Microsoft even uses it for [TypeScript](https://github.com/microsoft/TypeScript/blob/master/.eslintrc.json) and [VS Code](https://github.com/microsoft/vscode/blob/master/.eslintrc.json)).

Some examples of errors and warnings caught by ESLint in this repo: several unused variables, [the use of `String` instead of `string` as a type](https://github.com/wanasit/chrono/blob/9322a70e92f8e1bfe152232fa846d17070cf6812/src/common/parsers/AbstractTimeExpressionParser.ts#L6) (they're not equivalent!), [unused imports](https://github.com/wanasit/chrono/blob/9322a70e92f8e1bfe152232fa846d17070cf6812/src/locales/en/parsers/ENWeekdayParser.ts#L4) (potentially bloating bundle size when tree-shaking), unneeded escape characters in regular expressions. 

Currently this PR does not fix any of the warnings raised by ESLint. If adding ESLint seems like a good idea, I can add those fixes in additional commits. For the same reason, this PR also does not add the `eslint` npm command to the Husky precommit hook.


wanasit:
Hello @lostfictions. Thanks for working on this! I'm happy to include ESLint!

However, could you please help a little bit more:
* I prefer embedding eslintConfig in `package.json`, rather than having a separate `.eslintrc`. 
* It also would be nice if you could fix all `error` level issues (to not block commit hook).

lostfictions:
@wanasit ok! I've cleaned up errors and warnings. I still haven't added `npm run eslint` to the husky hook yet, but it should be ok to do so now.

wanasit:
Cool! thank you very much! @lostfictions

Issue #364, opened 2020-12-08
Timezone-induced test failures on CI


lostfictions:
When updating #361, I experienced an odd issue -- a few tests were failing on master (not my own branch!) for no clear reason. Here's the summary:

> Summary of all failing tests
>  FAIL  test/en/en_relative.test.ts (9.746 s)
>   ● Test - Relative date components' certainty
> 
>     expect(received).toBe(expected) // Object.is equality
> 
>     Expected: -240
>     Received: -300
> 
>       155 |         expect(result.start.get("day")).toBe(7);
>       156 |         expect(result.start.get("hour")).toBe(12);
>     > 157 |         expect(result.start.get("timezoneOffset")).toBe(refDate.getTimezoneOffset() * -1);
>           |                                                    ^
>       158 | 
>       159 |         expect(result.start.isCertain("year")).toBe(true);
>       160 |         expect(result.start.isCertain("month")).toBe(true);
> 
>       at test/en/en_relative.test.ts:157:52
>       at Object.<anonymous> (test/en/en_relative.test.ts:151:5)
> 
>   ● Test - Relative date components' certainty and imply timezone
> 
>     Expected date to be: Mon Nov 30 2020 11:00:00 GMT-0500 (GMT-05:00) Received: Sun Nov 29 2020 11:00:00 GMT-0500 (GMT-05:00) ([ParsingResult {index: 0, text: 'tomorrow at 5pm', ...}])
> 
>       186 |         result.start.imply("timezoneOffset", 60);
>       187 | 
>     > 188 |         expect(result).toBeDate(new Date("Sun Dec 1 2020 1:00:00 GMT+0900 (Japan Standard Time)"));
>           |                        ^
>       189 |         expect(result).toBeDate(new Date("Sun Nov 30 2020 17:00:00 GMT+0100"));
>       190 |     }
>       191 | 
> 
>       at Object.<anonymous> (test/en/en_relative.test.ts:188:24)
> 

This is current `master`, which is at commit [`6daa27d`](https://github.com/wanasit/chrono/commit/6daa27d0fae43a12c42d43c1bab2d79d1d4ed570). I'm on Node `v14.15.1`.

Most oddly, I rebased PR #361 on top of that commit, and didn't make any changes that should affect tests, but they pass on Travis instead of failing.

benaubin:
Hmm... feel like this could be related to timezones

ptz0n:
Yep, fixed in #368.

lostfictions:
[...]

PR #365, opened 2020-12-08
Update Travis Node versions

lostfictions:
[Node 11 and 13 are EOL](https://nodejs.org/en/about/releases/), so this PR updates the tested versions. ~~(I'm also curious if this will also make the test failure I'm seeing in #364 reproducible in CI.)~~ CI for this PR has passed, so the Node version does not appear related to the error in #364.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: node_js
 node_js:
-  - "11"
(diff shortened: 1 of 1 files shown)


benaubin (approved):
lgtm

PR #372, opened 2021-01-27
Make "in|within|for" optional if forwardDate=true

### LINKED ISSUE #371: chrono.parse("N months", new Date()) gives null
Not sure if it's expected.

I will try to provide a failing test, but for now the test-suite is partially failing on my machine, so it might take time.

```typescript
import * as chrono from 'chrono-node'
chrono.parseDate("2 months", new Date()) // returns null
```

flpvsk:
This let's us parse expressions like "give it 2 months" given that
forwardDate is set to true. At the same time if forwardDate is not set,
parsing fails as ambiguous.

Fixes #371

### DIFF
--- a/src/locales/en/parsers/ENTimeUnitWithinFormatParser.ts
+++ b/src/locales/en/parsers/ENTimeUnitWithinFormatParser.ts
@@ -4,9 +4,11 @@ import { ParsingComponents } from "../../../results";
 import { AbstractParserWithWordBoundaryChecking } from "../../../common/parsers/AbstractParserWithWordBoundary";
 
 export default class ENTimeUnitWithinFormatParser extends AbstractParserWithWordBoundaryChecking {
-    innerPattern(): RegExp {
+    innerPattern(context: ParsingContext): RegExp {
+        const prefix = context.option.forwardDate ? "" : "(?:within|in|for)\\s*";
         return new RegExp(
-            `(?:within|in|for)\\s*(?:(?:about|around|roughly|approximately|just)\\s*(?:~\\s*)?)?` +
+            prefix +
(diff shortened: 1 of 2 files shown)


src/locales/en/parsers/ENTimeUnitWithinFormatParser.ts
```diff
@@ -4,9 +4,11 @@ import { ParsingComponents } from "../../../results";
 import { AbstractParserWithWordBoundaryChecking } from "../../../common/parsers/AbstractParserWithWordBoundary";
 
 export default class ENTimeUnitWithinFormatParser extends AbstractParserWithWordBoundaryChecking {
-    innerPattern(): RegExp {
+    innerPattern(context: ParsingContext): RegExp {
+        const keywordOptional = context.option.forwardDate ? "?" : "";
         return new RegExp(
-            `(?:within|in|for)\\s*(?:(?:about|around|roughly|approximately|just)\\s*(?:~\\s*)?)?` +
+            `(?:within|in|for)${keywordOptional}` +
```
wanasit, line 10: I found appending the optional sign to the regex is more difficult to understand what will happens.

Could you please write the full question-marked prefix with you condition? 
```
const prefix= context.option.forwardDate ? ... : ... ;
return new RegExp(prefix + ...)
```

or just inline the whole thing.
```
return new RegExp(
  context.option.forwardDate ? `...` : `...` +
  ...
)
```

wanasit (commented):
Sorry for my slow reply. I add one small comment. Could you please update that?

src/locales/en/parsers/ENTimeUnitWithinFormatParser.ts
flpvsk: Fixed 👍

PR #373, opened 2021-02-03
Fixed and extend NL locale + add test for all

JeroenVdb:
I wanted to use this library for the `nl` locale, which was added via PR https://github.com/wanasit/chrono/pull/351. I noticed some parsers where not correct and/or incomplete and missing tests.

Changed/fixed
- Fixed NLMonthNameMiddleEndianParser: "januari 13, 2012" -> "13 januari 2012"

Added:
- Tests for all nl parsers and refiners
- Added NLCasualDateTimeParser for combined words like "morgennamiddag", "vanavond" and "gisterenochtend"
- Added NLCasualYearMonthDayParser for YMD pattern support


JeroenVdb:
@Pruxis maybe you should also check these changes.
I think the biggest (breaking) change is the format fix in NLMonthNameMiddleEndianParser.ts - but in my opinion the old format (January 13, 2012) was just wrong.

JeroenVdb:
Same for @Thymo as you added the implementation in version 1.

wanasit:
Wow! Thanks for the improvement!
Sorry for my slow reply.

PR #388, opened 2021-04-04
Fix timezones ParsingOption type

zuhairp:
Previously, the timezones parameter was expecting a a single key with the literal value "string" instead of key-value override pairs


Saadnajmi (approved):

wanasit:
Thanks for the fix!

PR #398, opened 2021-06-15
Added more support for Dutch.

jebbench:
I've tried to add more support for the Dutch language (I'm not a native speaker so it's based on Google translate and some help from a client).

Since rebasing off master today all of the tests have broken - they seem to be broken in master so I don't think it's anything I've done.


wanasit:
Nice work! Thanks for the improvement.

PR #411, opened 2021-10-15
Support empty (null) timezone in parser reference date

### LINKED ISSUE #407: Possible bug: 'now' is parsed with a known timezone offset
This could be a misunderstanding of the library on my behalf; apologies if so.

Using `chrono-node` v2.3.2 and parsing input where the word `now` is identified as a timestamp in the text, the `knownValues` for that timestamp suggest that the result has a known timezone (system local, I think) associated with it.

That seemed surprising and I wondered whether it might be a bug.  The relevant code is here: https://github.com/wanasit/chrono/blob/da9bd4a8c06feb9f2700e6e34d052fc4cc064b63/src/common/casualReferences.ts#L11

jayaddison-collabora:
Previously, any `timezone` value passed in the `reference` parsing context would always result in a non-empty `ReferenceWithTimezone.timezoneOffset`.  In particular, passing `null` would result in a timezone offset of `-0`.

Although there is probably a neater and more maintainable implementation possible, the intent here is to allow the caller to provide a `null` value for `timezone` -- the resulting `timezoneOffset` will only be overridden in cases when `chrono` finds a known/implied timezone in the text.

This is useful in situations where source texts do not have easily-determined timezones, and allows the caller to ignore results that 'look like' times according to `chrono` but where a definite (`isCertain`) timezone is not available.

**NB**: The diff here may be easier to view at this URL: https://github.com/jayaddison-collabora/chrono/compare/refactor/migrate-createRelativeFromRefInstant...jayaddison-collabora:issue-407/allow-empty-reference-timezone since it builds upon a refactor branch.

Resolves #407.

Relates to https://github.com/mattermost/mattermost-plugin-walltime/pull/40.

### DIFF
--- a/src/timezone.ts
+++ b/src/timezone.ts
@@ -193,6 +193,10 @@ export const TIMEZONE_ABBR_MAP = {
 };
 
 export function toTimezoneOffset(timezoneInput: string | number): number {
+    if (timezoneInput === null) {
+        return null;
+    }
+
     if (typeof timezoneInput === "number") {
         return timezoneInput;
     }
--- a/src/timezone.ts
+++ b/src/timezone.ts
@@ -192,7 +192,7 @@ export const TIMEZONE_ABBR_MAP = {
     YEKT: 360,
 };
 
-export function toTimezoneOffset(timezoneInput: string | number): number {
+export function toTimezoneOffset(timezoneInput: string | number): number | null {
     if (timezoneInput === null) {
         return null;
     }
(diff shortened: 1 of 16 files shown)


src/timezone.ts
```diff
@@ -193,6 +193,10 @@ export const TIMEZONE_ABBR_MAP = {
 };
 
 export function toTimezoneOffset(timezoneInput: string | number): number {
```
wanasit, line 195: Nit: `toTimezoneOffset(...): number?` or `toTimezoneOffset(...): number | null`

wanasit (commented):
Thanks for pointing out the problem.
I also like createRelativeFromRefInstant() -> createRelativeFromReference().

Could you please take a my small comment?

jayaddison-collabora:
Do you know when the next release for the `chrono` library might happen @wanasit?

wanasit:
ops. I really sorry for the delay. Publishing the new version now.

wanasit:
v2.3.3

jayaddison-collabora:
No problem, releases are ready when they are ready :)  Thank you!

PR #415, opened 2021-11-23
Fixup for #411: allow the 'now' casual reference function to support null reference timezones

jayaddison-collabora:
Apologies @wanasit - this bug/gap is something I missed during #411, due to not adding a relevant test case for parsing of text containing the phrase `now`.

The fix is to add the same guard statements around in the `now` function as were [added to the `createRelativeFromReference` function](https://github.com/wanasit/chrono/pull/411/files#diff-1c0edb2e88e4f027868f0b0389e98b0ee10cf56d0949c931f9b5f2c7831bc191L184-R190).

This relates to #407.

### DIFF
--- a/src/common/casualReferences.ts
+++ b/src/common/casualReferences.ts
@@ -8,7 +8,9 @@ export function now(reference: ReferenceWithTimezone): ParsingComponents {
     const component = new ParsingComponents(reference, {});
     assignSimilarDate(component, targetDate);
     assignSimilarTime(component, targetDate);
-    component.assign("timezoneOffset", targetDate.utcOffset());
+    if (reference.timezoneOffset !== null) {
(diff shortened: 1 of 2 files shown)


wanasit:
@jayaddison-collabora fyi. there is also this additional fix to make it works 
801ca4e

jayaddison-collabora:
> @jayaddison-collabora fyi. there is also this additional fix to make it works 801ca4e

Hrm - I can't find that commit.  Is it on a branch?

wanasit:
https://github.com/wanasit/chrono/commit/801ca4ee2985b1249f5b83aceadef7d0e8b194cb

jayaddison-collabora:
Brilliant, thank you!

PR #417, opened 2021-12-12
Migration zh-hant from v1

DingWeizhe:
Move ZH-Hant from v1

### DIFF
--- a/src/common/parsers/AbstractParserWithWordBoundary.ts
+++ b/src/common/parsers/AbstractParserWithWordBoundary.ts
@@ -27,7 +27,7 @@ export abstract class AbstractParserWithWordBoundaryChecking implements Parser {
     }
 
     extract(context: ParsingContext, match: RegExpMatchArray) {
-        const header = match[1];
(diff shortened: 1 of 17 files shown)


DingWeizhe:
I have temporarily released the new package [(chrono-node-v2-zh-preview)](https://www.npmjs.com/package/chrono-node).
If you need to use it first, you can try it.
But when this PR is approved, I will set the package as deprecated.

wanasit:
@DingWeizhe I'm very sorry for my slow reply. 
Thank you for you improvement! I'm really appreciate you help working on this.

wanasit:
Your change looks good to me. The only comment/concern I have is I'm not sure if `zhHant` (e.g. `chrono.zhHant`) is a common way to refer to the locale (e.g. I thought it should be `chrono.zh` and `chrono.zh.hant`)

I'm merging your change, but let me do some research about what is the common pattern for that first before publishing this.

wanasit:
Also. Happy Christmas! btw.

Issue #360, opened 2020-11-14
Aggressive timezone parsing produces many incorrect results


lostfictions:
[`ExtractTimezoneAbbrRefiner`](https://github.com/wanasit/chrono/blob/master/src/common/refiners/ExtractTimezoneAbbrRefiner.ts) interprets valid words in many languages as timezones, leading to incorrect parsing.

Some examples:

```
> chrono.parseDate('in 1 hour eggs and milk')
2020-11-14T02:35:28.000Z   -- correct
> chrono.parseDate('in 1 hour get eggs and milk')
2020-11-13T17:35:32.000Z   -- "get" interpreted as timezone
> chrono.parseDate('in 1 hour eat dinner with john')
2020-11-13T18:38:42.000Z   -- "eat" interpreted as timezone 
> chrono.parseDate('in 1 hour wet clothes into dryer')
2020-11-13T21:39:20.000Z   -- "wet" interpreted as timezone
```
Other languages are also subject to errors:

```
> chrono.fr.parseDate('dans 1 heure mon anniversaire')
2020-11-14T02:40:32.000Z   -- no problem
> chrono.fr.parseDate('dans 1 heure cest mon anniversaire')  -- eg. casual orthography in french
2020-11-13T19:40:50.000Z   -- "cest" interpreted as timezone
```

As a first suggestion, I think timezone parsing should definitely not be applied to relative times. As far as I know, regardless of language it never makes sense to apply a timezone to a relative time -- "in one hour pacific time" is not a thing people can mean; one hour just means one hour.

I would happily submit a PR for this, but unfortunately it looks tricky to achieve given the current structure of the code, where all refiners are always applied regardless of parser. It seems like this would need to change for this fix to be achievable, but I think it's important -- otherwise many sentences will continue to be incorrectly parsed. How would you like to proceed?

---

Beyond the obvious scenario of relative times described above, I think this issue can still arise even when specifying absolute times -- and the solution may be more tricky. As a simple example, "At 1pm get eggs and milk" is a similar case where parsing would be incorrectly applied. Here are some potential solutions I can think of:

- Only parse uppercase timezones. For example, "at 1pm GET eggs and milk" would be parsed in the GET timezone but "1pm get" would not. This seems potentially brittle -- there are many reasons input might be in uppercase, so it may not reliably disambiguate correctly.
- Make timezone parsing opt-in. This would ensure no ambiguities are possible, but would require the developer to understand the tradeoffs based on their expected use case.
[...]

PR #425, opened 2022-01-24
Address handling of ambiguous timezone references

### LINKED ISSUE #424: Interpretation of ambiguous tokens as timezone names
Upstreaming this issue from mattermost/mattermost-plugin-walltime#53 - it looks like the [cause](https://github.com/mattermost/mattermost-plugin-walltime/issues/53#issuecomment-1019992221) was an upgrade to `chrono` from v1.3.11 to v2.3.4.

Certain text tokens (words) may be misinterpreted as timezones by the plugin under some circumstances.

More precisely, an issue was experienced and reported against [Collabora's currently-deployed version of the plugin](https://gitlab.collabora.com/tools/mattermost/mattermost-plugin-walltime/-/tree/d00bde4efd968d29bcc4854dcd188a292867f264).

In particular, the phrase:

```
People visiting Buñol towards the end of August get a good chance to participate in La Tomatina (under normal circumstances)
```

... was reformatted to ...

```
People visiting Buñol towards the end of `August get` *(Thu, Aug 1, 2019 1:00 AM PDT)* a good chance to participate in La Tomatina (under normal circumstances)
```

This is due to the token `get` being misinterpreted by the plugin as [`GET` (Georgia Standard Time)](https://time.is/GET).

jayaddison-collabora:
#### Summary
Aims to address incorrect interpretation of ambiguous tokens (for example, the word `get` following a month name) as timezones by the library.

#### Ticket Link
Resolves #424.

### DIFF
--- a/test/en/negative_cases.test.ts
+++ b/test/en/negative_cases.test.ts
@@ -67,3 +67,10 @@ test("Test - Date with version number pattern", () => {
         expect(result.text).toBe("2015-09-24");
     });
 });
+
+test("Test - Month with ambiguous trailing token", () => {
+    testUnexpectedResult(
+        chrono,
+        "People visiting Buñol towards the end of August get a good chance to participate in La Tomatina (under normal circumstances)"
+    );
+});


jayaddison-collabora:
@wanasit Could you help with a fix for this issue?  I'm not exactly how best to approach this ambiguity in the parser.

jayaddison-collabora:
@wanasit change of plan; I'm going to try to determine a fix for this myself.  This problem has appeared in the context of my employment @collabora, so I (and we, the company) should not expect any free work input from you.  Sorry for my previous comment, which may have appeared entitled and demanding.  I'll try to determine a fix for #424 here soon.

jayaddison-collabora:
From further research: one possibility here is to extend the filtering in [`UnlikelyFormatFilter.ts`](https://github.com/wanasit/chrono/blob/1320d7c403cb4da23a8e06d82fcff5748f552936/src/common/refiners/UnlikelyFormatFilter.ts).

That code is generally composed of filters that apply under casual formatting conditions, and separate, more stringent filters that apply during parsing in 'strict' mode.

It would be possible to place a check for 'month and timezone known, but not day' within one of those two areas, however this could be disruptive for users of the `chrono` library.

Perhaps it makes sense not to handle this within `chrono` yet, but to trial a potential solution within the `mattermost-plugin-walltime` plugin codebase and learn from that.

wanasit:
@jayaddison-collabora Hello. I'm really sorry about my slow reply. 

I have checked your proposed solution, but decided to fix it differently in  afbba587c6983222c152526453bb7513506f0fa6. Updating the library (e.g. v2.3.6+) should fix your specific case (note that the "August" would still be extracted, which differ from the test you wrote here)

The issue about the timezone abbreviation was already discussed once in https://github.com/wanasit/chrono/issues/360. I agree that the extraction is too aggressive that it leads too false negatives, but updating that would break many tests. I would save that for the next coming minor update.

jayaddison-collabora:
Thank you @wanasit!  I had a sense that you might be able to think of a more maintainable / preferable solution.  I'll upgrade to 2.3.6 to resolve this application-side soon.

PR #426, opened 2022-01-25
Update README.md about ParsedResult.length attribute

jonathanberger:
Only by reading the source did I figure out that ParsedResult.length existed, and that if it was set to zero I should treat the parse as having failed. This is the behavior of chrono.parseDate.

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,7 +118,8 @@ chrono.parseDate('Friday', referenceDate, { forwardDate: true });
 
 #### ParsedResult
 * `refDate: Date` The [reference date](#reference-date) of this result
-* `index: number` The location within the input text of this result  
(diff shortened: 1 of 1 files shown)


wanasit:
Hello. @jonathanberger. Maybe the documentation is not clear, but your update is still not correct:

* The `parse()` function returns an array `ParsedResult[]`, not one ParsedResult. And if the array is empty (results.length == 0), this input doesn't have dates.
* Each `ParsedResult` object itself doesn't have length attribute.

jonathanberger:
Ah, sorry and thank you for the message. I see now the typescript definition of `parse` in the docs that I missed before. Perhaps if you hear similar confusion again if might be worth clarifying.

PR #427, opened 2022-01-29
feat: chinese simplified support

QingWei-Li:
Modified from Traditional Chinese(zh-hant), the two languages are mainly the difference between words and expressions. The rules are the same.


wanasit:
Hey @QingWei-Li. Very cool! thanks for working on this!

Issue #380, opened 2021-03-03
How to improve parsing performance?


geo-systems:
Hi there,

I need to write a program, which parses many dates in different formats.

I've encountered a performance issue with `chrono`. Here's a benchmark, which demonstrates the issues:

```js
const chrono = require("chrono-node")
let max = 100000;

console.time("chrono parsing");
for (let i = 0; i < max; i++) {
  chrono.en.strict.parseDate("11/12/2021");
}
console.timeEnd("chrono parsing");

console.time("native parsing");
for (let i = 0; i < max; i++) {
  new Date("11/12/2021");
}
console.timeEnd("native parsing");
``` 

The output is:

```
chrono parsing: 34.235s
native parsing: 62.26ms
```

I understand that `chrono` does much more than the native parsing and it's understandable that it's slower. However 500 times slower seems a bit too much.

Is there a way to speed up `chrono`? Can I disable some patterns or locales to speed it up?

wanasit:
Hello. Thanks for reporting this. This could be improved by reducing the backtracking as in #382, but I afraid it won't be as fast as native javascript.

> However 500 times slower seems a bit too much.

I am afraid it is not. I believe [the Date constructor is native](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toSource) (imagine optimized C/C++ routine built into Javascript engine) while the Chrono parsing is a normal Javascript code run on top of the engine.

Do you have any specific parsing latency number to achive or any time-critical use-case you have in mind?

geo-systems:
Thanks for looking into this!

For 100K records (as in the benchmark), up to 10 seconds would be acceptable. I'm parsing a few million records and with the current performance the user would have to wait 20-30 minutes.

I've also conducted benchmark with another library called [Sugar JS](https://sugarjs.com/):

```js
const chrono = require("chrono-node")
const sugar = require("sugar")

let max = 100000;

console.time("chrono parsing");
for (let i = 0; i < max; i++) {
  chrono.en.strict.parseDate("11/12/2021");
}
console.timeEnd("chrono parsing");

console.time("sugar parsing");
for (let i = 0; i < max; i++) {
  sugar.Date.create('11/11/2021');
}
console.timeEnd("sugar parsing");

console.time("native parsing");
for (let i = 0; i < max; i++) {
  new Date("11/12/2021");
}
console.timeEnd("native parsing");
```

The result shows that Sugar JS is about 15 times faster:

```
[...]

PR #429, opened 2022-02-04
feat: Add support for custom weekStart

liamcain:
## Summary

```js
const chrono = new Chrono();
chrono.parseDate("this Sunday", new Date(), { 
  locale: { weekStart: 1 }
});
```

### Use case

Parsing `this <day_of_week` always assumes the week starts on Sunday. For locales that start their week on Monday, this means the wrong date will be implied. e.g.`this Sunday` returns 31 January instead of 06 February.

### Solution

I opted to allow passing in a dayjs locale so that a user can provide any `weekStart` that they want with minimal changes. It does have a fallback to the `en` locale which isn't exactly _correct_ but maintains parity with the current version.

## Tasks

- [x] Add test coverage
- [x] Update README

~~## Questions~~

- ~~Locally a lot of tests are failing for me (even on `master`). This seems to be related to the custom jest matcher `toBeDate` not handling daylight savings. Any ideas for a workaround?~~ Fixed by changing jest timezone to UTC.

### DIFF
--- a/test/en/en_timezone_exp.test.ts
+++ b/test/en/en_timezone_exp.test.ts
@@ -97,7 +97,7 @@ test("Test - Not parsing timezone from relative time", function () {
         const expectedInstant = new Date("Sun Nov 29 2020 14:24:13 GMT+0900 (Japan Standard Time)");
 
         testSingleCase(chrono, "in 1 hour GMT", refInstant, (result, text) => {
-            expect(result.text).toBe("in 1 hour");
+            expect(result.text).toBe("in 1 hour GMT");
             expect(result.start).toBeDate(expectedInstant);
         });
     }
@@ -107,7 +107,7 @@ test("Test - Not parsing timezone from relative time", function () {
         const expectedInstant = new Date("Sun Nov 29 2020 14:24:13 GMT+0900 (Japan Standard Time)");
 
         testSingleCase(chrono, "in 1 hour GMT", { instant: refInstant, timezone: "JST" }, (result, text) => {
-            expect(result.text).toBe("in 1 hour");
+            expect(result.text).toBe("in 1 hour GMT");
             expect(result.start).toBeDate(expectedInstant);
         });
     }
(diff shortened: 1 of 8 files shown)


liamcain:
Okay, there was a simple fix to my issue with running the tests locally.

```bash
TZ=UTC yarn run test
```

test/en/en_timezone_exp.test.ts
```diff
@@ -97,7 +97,7 @@ test("Test - Not parsing timezone from relative time", function () {
         const expectedInstant = new Date("Sun Nov 29 2020 14:24:13 GMT+0900 (Japan Standard Time)");
 
         testSingleCase(chrono, "in 1 hour GMT", refInstant, (result, text) => {
-            expect(result.text).toBe("in 1 hour");
+            expect(result.text).toBe("in 1 hour GMT");
```
liamcain, line 100: Sorry, these changes are unrelated but the tests were failing locally for me.

wanasit:
Hello @liamcain. Thanks for working on this. 

I understand the feature you request (I also like your external API design). However, the implementation direction is too different from what I have in mind for Chrono.

First, Chrono is moving away from `dayjs`. We have been using dayjs just for calendar-based calculation, but it is not necessary and it would be more efficient and reliable when replaced with Native Javascript (e.g. https://github.com/wanasit/chrono/commit/cdfb6bcb82d3df51c3a4e50ec1a71e2848edc68d, ref https://github.com/wanasit/chrono/issues/380). 

(Note that the usage "en" locale as a default, also break the timezone offset calculation is also an unacceptable side effect to the current Chrono users)

So, instead of using and exposing dayjs and its locale, we should just accept weekStart as a parsing option (not that the example in your readme doesn't have locale wrapped): 

```
chrono.parseDate('Sunday', referenceDate, { weekStart: 1 });
```

and then manually adjust the offset calculation on the functions in weeks util that you have already identified:

```
toDayJSClosestWeekday(refDate: Date, offset: number, weekStart: number) {
    ...
}

// or eventually toDateClosestWeekday, and skip DateJs completely.
```

wanasit:
I know what I request is not easy, but I still hope you could work on it (so, we could publish the feature to your full credit).
However, I could also make the change next week.

liamcain:
@wanasit, Hi! sorry for not getting back to you earlier, I got a bit caught up with some other responsibilities.

The changes you proposed make a lot of sense and I think I'm up for the task, provided you haven't started working on it at all.

wanasit:
Thank you. 

Yeah. I haven't worked on it... 😅 
(Sorry, my timeline or promise date as a weekend hobbyist opensource owner is totally not reliable.)

PR #435, opened 2022-02-15
[DE] Add optional "h" to time expression

georgd:
The time `4pm` is often represented as `16h` in German.


wanasit:
Thanks for your contribution here too. Could you also please add some test cases?

Issue #439, opened 2022-03-01
Tests fail when DST switch interferes


georgd:
This was originally reported in https://github.com/wanasit/chrono/pull/434 and partially fixed with https://github.com/wanasit/chrono/commit/e3b76090e75a95e92bc821ec72ddf367db66b8ac.

Some tests fail when the system is located in a timezone where a daylight saving time switch occurs between a reference or source date and the target date. The target date will be an hour off of what is expected. This is mathematically correct but breaks the tests. In most cases, it’s just the hour that’s off (and the timeZoneOffset) but in some cases it’s even the day (when the source time is midnight, like in https://github.com/wanasit/chrono/blob/61ae8a34ba1beb736d86f11d86a074f7568ba239/test/en/en_time_units_ago.test.ts#L174).

Is it safe to manipulate the tests to make them pass?

Here’s the list of the failing tests:

```
 Summary of all failing tests
 FAIL  test/en/en_time_units_within.test.ts (20.31 s)
  ● Test - The normal within expression

    Expected date to be: Sat Nov 10 2012 12:14:00 GMT+0100 (Mitteleuropäische Normalzeit) Received: Sat Nov 10 2012 11:14:00 GMT+0100 (Mitteleuropäische Normalzeit) ([ParsingComponents {knownValues: {"month":11,"year":2012}, impliedValues: {"day":10,"hour":12,"minute":14,"second":0,"millisecond":0,"timezoneOffset":120}}, reference: {"instant":"2012-08-10T10:14:00.000Z"}])

      106 |         expect(result.text).toBe("within a few months");
      107 |
    > 108 |         expect(result.start).toBeDate(new Date(2012, 10, 10, 12, 14));
          |                              ^
      109 |     });
      110 |
      111 |     testSingleCase(chrono, "within one year", new Date(2012, 7, 10, 12, 14), (result) => {

      at checkResult (test/en/en_time_units_within.test.ts:108:30)
      at Object.<anonymous> (test/en/en_time_units_within.test.ts:104:5)

 FAIL  test/en/en_relative.test.ts
  ● Test - Relative date components' certainty

    expect(received).toBe(expected) // Object.is equality

    Expected: 60
    Received: 120

      184 |         expect(result.start.get("day")).toBe(7);
      185 |         expect(result.start.get("hour")).toBe(12);
    > 186 |         expect(result.start.get("timezoneOffset")).toBe(-expectedDate.getTimezoneOffset());
          |                                                    ^
      187 |
[...]

PR #440, opened 2022-03-02
DE enhancements

georgd:


georgd:
Please note: I was blocked from commiting (even locally) by https://github.com/wanasit/chrono/issues/439. So, in order to be able to commit any changes, I disabled all non-de tests on my machine.

PR #442, opened 2022-03-10
Enhance de

georgd:


test/de/de_time_units_later.test.ts
```diff
@@ -0,0 +1,151 @@
...
+        expect(result.index).toBe(0);
+        expect(result.text).toBe("in 1,5 Stunden");
+        expect(result.start.get("hour")).toBe(14);
+        expect(result.start.get("minute")).toBe(10);
+
+        expect(result.start).toBeDate(new Date(2012, 7, 10, 14, 10));
+    });
+});
+
+// test("Test - Strict mode", function () {
```
wanasit, line 134: Please don't leave the commented out code like this.
If we don't have DE strict mode, let's remove it.

test/de/de_time_units_ago.test.ts
```diff
@@ -0,0 +1,332 @@
...
+        expect(result.start.get("day")).toBe(1);
+
+        expect(result.index).toBe(0);
+        expect(result.text).toBe("vor ein paar Tagen");
+
+        expect(result.start).toBeDate(new Date(2012, 8 - 1, 1));
+    });
+});
+
+//test("Test - Nested time ago", function () {
```
wanasit, line 301: Same as previous comment about not leaving commented code.

If we are not ready to support the nested time ref, lets remove this and add it later.

src/locales/de/parsers/DETimeUnitRelativeFormatParser.ts
```diff
@@ -5,35 +5,47 @@ import { AbstractParserWithWordBoundaryChecking } from "../../../common/parsers/
...
-        if (/vor/.test(modifier) || /letzte/.test(modifier) || /vergangen/.test(modifier)) {
+        let num = match[2] ? parseNumberPattern(match[2]) : 1;
+        if (/vorvorletzte/.test(modifier) || /überübernächste/.test(modifier)) {
+            num += 2;
+        } else if (/vorletzte/.test(modifier) || /übernächste/.test(modifier)) {
+            num += 1;
+        }
+
+        const unit = TIME_UNIT_DICTIONARY[match[4].toLowerCase()];
+        let timeUnits = {};
```
wanasit, line 40: Nit: first, I'd rename `num` to `value` and I think `let timeUnits = { unit: num };` is clearer to understand.

wanasit:
Hey @georgd. Thanks for the improvement!
Could you please address or reply my comments?

dpschen:
Hey @georgd, this is really useful! In case you currently don’t have time to continue on this would you be fine if I would pick up this PR?

georgd:
Hi, indeed I sadly won't have time for this for another couple of months, as it seems. I'd be happy to hand this over to you 😀

georgd:
Back then, I started refactoring the code to bring the DE part on par with EN but I am better with regex than with js so didn't get far in the little time I had. When I find the time, would you mind me proposing improvements?

PR #444, opened 2022-03-25
Fix bug: Nothing happens to the original object after manipulation

delapecci:
Fix bug: Nothing happens to the original dayjs object after manipulations, should assign to the new cloned one


wanasit:
Thank you @delapecci.
I am curious why didn't those fail the tests. Could you also add test cases to confirm the issue got addressed?

PR #448, opened 2022-04-16
Add ESM build output in addition to CJS

JoakimNil:
Output to CommonJS in dist as usual, but add a separate build output in dist/esm that targets ES modules.

My motivation for this change is to optimize performance for the browser. Most importantly this change enables advanced tree-shaking, reducing the size of chrono in bundles by ~80%.

I made an effort to change as little as possible to avoid breaking changes—dependents are now free to choose whether to use the CJS or ESM, and the changes to package.json ensures that most tooling will automatically choose the appropriate version.
There is an argument to be made to switch entirely to ESM, but this would be a breaking change that would make it difficult for CJS packages to use chrono as a dependency. So it is better to target both IMO.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,18 @@
     "test": "./test"
   },
   "main": "dist/index.js",
-  "types": "dist/index.d.ts",
+  "module": "dist/esm/index.js",
+  "exports": {
+    ".": {
+      "require": "./dist/index.js",
+      "import": "./dist/esm/index.js"
+    }
+  },
+  "engines": {
+    "node": "^12.20.0 || ^14.13.1 || >=16.0.0"
+  },
+  "sideEffects": false,
+  "types": "dist/esm/index.d.ts",
   "devDependencies": {
     "@types/jest": "^26.0.4",
     "@typescript-eslint/eslint-plugin": "^4.7.0",
@@ -32,9 +43,12 @@
     "husky": "^4.2.5",
     "jest": "^26.1.0",
     "prettier": "^2.0.5",
+    "run-script-os": "^1.1.6",
     "ts-jest": "^26.1.1",
+    "tsc-esm-fix": "^2.12.4",
     "typedoc": "^0.20.36",
-    "typescript": "~4.2.4"
+    "typescript": "~4.2.4",
+    "rimraf": "^3.0.2"
   },
   "husky": {
     "hooks": {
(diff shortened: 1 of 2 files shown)


JoakimNil:
I should add that it's difficult to test all dev environments to ensure that nothing breaks for anyone. Specifically, some tooling may not work correctly with e.g. the new `module` or `exports` fields in `package.json`, even though all existing fields (except `types` path) were left unchanged. So it would be an advantage to have several users test this change in different environments. And this should at least be a semver minor version bumb.

wanasit:
Hey @JoakimNil thanks for working on this, but I would like to take time to research and think about this. Sorry I'm also not very good at the Javascript build and distribute options. 

Do you mind share the resources, references, or examples project setup where I can learn from?

jamauro:
> My motivation for this change is to optimize performance for the browser. Most importantly this change enables advanced tree-shaking, reducing the size of chrono in bundles by ~80%.

Would be great to see this. Any update on this PR?

odinho:
This would be absolutely awesome to have. Dragging along all the locales is quite un-nice when they're not used. They should be possible to lazy-load when they're needed instead :)

A primer to get into the entrypoints (main, module, exports) in `package.json` is the node page about it: https://nodejs.org/api/packages.html

This PR looks good to me.

JoakimNil:
Sorry @wanasit, life got in the way and I kind of forgot about this. Here are a few key points:
* There are two main module systems, CommonJS/CJS, and ECMAScript/ESM modules.
* CJS is the legacy system, and ESM is newer, with quite a few benefits, like tree-shaking support.
* These module systems don't work very well together: importing ESM modules in a CJS project is difficult.
* Any NPM package can target either CJS or ESM, or both.
* This PR doesn't change the existing CJS Chrono stuff, it just adds ESM output that makes it possible for users to specifically import ESM rather than the default CJS Chrono.

Here are some additional resources for learning about this:
* https://nodejs.org/api/esm.html
* [A huge thread on how sindresorhus did this](https://github.com/sindresorhus/meta/discussions/15)
* The key npm module used to achieve this, tsc-esm-fix has a very good and relevant [readme](https://www.npmjs.com/package/tsc-esm-fix).

One last point I should mention is that this PR lets tooling determine the module system by using the "type" setting in the nearest package.json, in accordance with the [node.js docs](https://nodejs.org/api/packages.html#determining-module-system).

wanasit:
Thanks for the reference. I'll try to merge this today.

PR #449, opened 2022-04-16
Parse timezone offsets that are enclosed in parentheses; restrict max tzOffset

JoakimNil:
Simple improvement. In making this change, I made a few trade-offs:

* It allows text with an open parenthesis without a closing parenthesis, e.g. `10.30 (UTC+4`. I don't see a huge benefit to disallowing this, and doing so would make the regex much more complex.
* It requires a GMT/UTC prefix to match offsets wrapped in parentheses, i.e. allow `10.30 (UTC+4)`, disallow `10.30 (+4)`. I did this because I have found chrono to be too eager to match math-like expressions. This may, however, be improved in 4756d0dd5b816d3e5aa3b90024e064750ce82f98, so I'm not sure if I have made the right call. On the other hand, I think the `(+4)` format is very rarely used.


JoakimNil:
I also limited the offsets that can be matched in ExtractTimezoneOffsetRefiner to max 14, since no timezones has higher offsets.

wanasit:
Cool! Thanks for the improvement.

PR #454, opened 2022-05-21
fix(README): ParsingReference "instant" instead of "instance"

am1ru1:
As per https://github.com/wanasit/chrono/blob/c4cd375591f1c51c5e4001814203bda5a838bbdd/src/chrono.ts#L151, it should be "instant" ParsingReference instead of "instance"


wanasit:
Ops. Thank you so much.

PR #457, opened 2022-06-17
[FR] small fix for duration

trothe:
Just a small fix for yet another way to specify a duration in French (e.g. régler une minuterie **de** 5 minutes)


wanasit:
Thanks for the fix.

PR #458, opened 2022-06-26
Adds Russian locale

mikemikhaylov:
Hey @wanasit as we discussed a few weeks ago, here is the PR with Russian language support.
Also there some changes in "common" codebase. Here is the list and motivation: 

1.  casualReferences.ts - I added a few methods for casual phrases, like "the day after tomorrow" (ru locale has single word definition for it, so I decided to support it) etc. This file looked like a right place for this.
2.  AbstractParserWithWordBoundary.ts - to support Cyrillic analog for \W  I needed to add **u** to RegEx flags and change some common parts of regular expression. Without it it considered Cyrillic letters as non-word characters. I didn't want to affect other locales in any way, at least in this PR since it's big enough on it's own, so I just added a method to this base class with default pattern left boundary and overrode in my class.
3. AbstractTimeExpressionParser.ts same thing though it was bit trickier. Also I removed escaping for **\\:|\\：** since it caused an error with unicode regex and didn't seem necessary in this context.

If you have any feedback on this PR would be happy to hear!

### DIFF
--- a/src/common/casualReferences.ts
+++ b/src/common/casualReferences.ts
@@ -1,6 +1,6 @@
 import { ParsingComponents, ReferenceWithTimezone } from "../results";
 import dayjs from "dayjs";
-import { assignSimilarDate, assignSimilarTime, assignTheNextDay, implySimilarTime } from "../utils/dayjs";
+import { assignSimilarDate, assignSimilarTime, implySimilarTime } from "../utils/dayjs";
 import { Meridiem } from "../index";
 
 export function now(reference: ReferenceWithTimezone): ParsingComponents {
@@ -26,37 +26,24 @@ export function today(reference: ReferenceWithTimezone): ParsingComponents {
  * The previous day. Imply the same time.
  */
 export function yesterday(reference: ReferenceWithTimezone): ParsingComponents {
-    let targetDate = dayjs(reference.instant);
-    const component = new ParsingComponents(reference, {});
-    targetDate = targetDate.add(-1, "day");
-    assignSimilarDate(component, targetDate);
-    implySimilarTime(component, targetDate);
-    return component;
+    return theDayBefore(reference, 1);
 }
 
-export function theDayBeforeYesterday(reference: ReferenceWithTimezone): ParsingComponents {
-    let targetDate = dayjs(reference.instant);
-    const component = new ParsingComponents(reference, {});
-    targetDate = targetDate.add(-2, "day");
-    assignSimilarDate(component, targetDate);
-    implySimilarTime(component, targetDate);
-    return component;
+export function theDayBefore(reference: ReferenceWithTimezone, numDay: number): ParsingComponents {
+    return theDayAfter(reference, -numDay);
 }
 
 /**
  * The following day with dayjs.assignTheNextDay()
  */
 export function tomorrow(reference: ReferenceWithTimezone): ParsingComponents {
-    const targetDate = dayjs(reference.instant);
-    const component = new ParsingComponents(reference, {});
-    assignTheNextDay(component, targetDate);
-    return component;
+    return theDayAfter(reference, 1);
 }
 
-export function theDayAfterTomorrow(reference: ReferenceWithTimezone): ParsingComponents {
+export function theDayAfter(reference: ReferenceWithTimezone, nDays: number): ParsingComponents {
     let targetDate = dayjs(reference.instant);
     const component = new ParsingComponents(reference, {});
-    targetDate = targetDate.add(2, "day");
+    targetDate = targetDate.add(nDays, "day");
     assignSimilarDate(component, targetDate);
     implySimilarTime(component, targetDate);
     return component;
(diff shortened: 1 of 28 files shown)


wanasit:
@mikemikhaylov Thanks for the contribution! 

Because your change is big, I'd need some time to review everything, but will try to give you feedback and merge this as soon as possible.

src/common/parsers/AbstractTimeExpressionParser.ts
```diff
@@ -3,23 +3,23 @@ import { ParsingComponents, ParsingResult } from "../../results";
 import { Meridiem } from "../../index";
 
 // prettier-ignore
-function primaryTimePattern(primaryPrefix: string, primarySuffix: string) {
+function primaryTimePattern(leftBoundary: string, primaryPrefix: string, primarySuffix: string, flags: string) {
     return new RegExp(
-        "(^|\\s|T|\\b)" +
+        leftBoundary +
```
wanasit, line 8: Could you follow the existing pattern and use the inline parameter format?

wanasit: Actually, if you look at more recent code, we have been using [template literals](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) than normal string, could you also help change everything here into template literals?

src/common/casualReferences.ts
```diff
@@ -34,6 +34,15 @@ export function yesterday(reference: ReferenceWithTimezone): ParsingComponents {
     return component;
 }
 
+export function theDayBeforeYesterday(reference: ReferenceWithTimezone): ParsingComponents {
```
wanasit, line 37: This function is limited to very narrow usage. To make it more flexible, I would make `theDayBefore(reference: ReferenceWithTimezone, numDay: number)`.

I would also update yesterday
```
yesterday(reference: ReferenceWithTimezone) {
 return theDayBefore(reference, 1);
} 
```

src/common/casualReferences.ts
```diff
@@ -44,6 +53,15 @@ export function tomorrow(reference: ReferenceWithTimezone): ParsingComponents {
     return component;
 }
 
+export function theDayAfterTomorrow(reference: ReferenceWithTimezone): ParsingComponents {
```
wanasit, line 56: Similarly, I would make it `theDayAfter(reference: ReferenceWithTimezone, nDays: number)`

src/common/casualReferences.ts
```diff
@@ -52,3 +70,53 @@ export function tonight(reference: ReferenceWithTimezone, implyHour = 22): Parsi
...
+    const component = new ParsingComponents(reference, {});
+    if (targetDate.hour() < 6) {
+        targetDate = targetDate.add(-1, "day");
+    }
+    assignSimilarDate(component, targetDate);
+    component.imply("hour", implyHour);
+    return component;
+}
+
+export function evening(reference: ReferenceWithTimezone, implyHour = 20): ParsingComponents {
```
wanasit, line 85: This function could be replaced with `tonight(ref, 20)`, right?

wanasit:
@mikemikhaylov The review is shorter than I thought. I think you did a great job of understanding and extending the existing abstraction. I still have a few comments, but overall looks good. 

Please consider my comments. I'll try to merge and release this on this weekend.

src/common/casualReferences.ts
mikemikhaylov: Actually I'm not sure about this. Because `tonight` has known date and `evening` doesn't. It's more like `morning`, which is just morning with unknown date.

src/common/casualReferences.ts
mikemikhaylov: Switched to theDayBefore and theDayAfter in https://github.com/wanasit/chrono/pull/458/commits/824f832bd610a2aaf531977cfce4a790431f7764

src/common/parsers/AbstractTimeExpressionParser.ts
mikemikhaylov: Can you please elaborate on this? 
You mean using template literals like here for all lines from 8 to 21?
https://github.com/wanasit/chrono/blob/4e02153671af4e2895154860a78126aa036f8566/src/locales/en/parsers/ENCasualYearMonthDayParser.ts#L14

wanasit: Yes. Let's use the template literals for every string that makes up any pattern. 

In addition to the consistency, it would make injecting/replacing with parameters more natural.

wanasit (approved):

src/locales/ru/constants.ts
```diff
@@ -0,0 +1,344 @@
...
+    четырех: 4,
+    четырёх: 4,
+    пять: 5,
+    пяти: 5,
+    шесть: 6,
+    шести: 6,
+    семь: 7,
+    семи: 7,
+    восемь: 8,
+    восемьми: 8,
```
spinybeast, line 128: typo?

mikemikhaylov: yes, you are right, thanks
fixed this and another one in "февряль" in https://github.com/wanasit/chrono/pull/461/files

PR #461, opened 2022-07-08
Fixes for ru locale

mikemikhaylov:
Fixes spelling of "February" and "eight", correct name for RUMergeDateRangeRefiner class


wanasit:
Thanks for the fix.

PR #462, opened 2022-07-25
ru: removes unnecessary replace in parseOrdinalNumberPattern

mikemikhaylov:
Hi, 
I noticed that in ru locale in `parseOrdinalNumberPattern` I used English suffixes. I guess it worked because parseInt just ignores non-digital ending.
Documentation:
> If parseInt encounters a character that is not a numeral in the specified radix, it ignores it and all succeeding characters and returns the integer value parsed up to that point.

I suggest to remove this `replace`.


wanasit:
Thanks for the fix.

>  I guess it worked because parseInt just ignores non-digital ending.
I am not sure if that is the reason. I think because that suffix removal is optional and most Russian text won't be affected anyway.

PR #466, opened 2022-08-21
Fix: Incorrect date(day) value in Chinese Local time expression parsers

maxwellyue:
Fix: Incorrect date(day) value in Chinese Local time expression parsers when date is not 'Today'.


wanasit:
@maxwellyue Thank you very much for the fix!

PR #473, opened 2022-10-07
IT fix+add tests

federicodilo:
*adding tests
*adding some time adverbs
*fixing some errors in other PR

### DIFF
--- a/src/locales/it/parsers/ITMonthNameMiddleEndianParser.ts
+++ b/src/locales/it/parsers/ITMonthNameMiddleEndianParser.ts
@@ -1,0 +1,78 @@
+import { ParsingContext } from "../../../chrono";
+import { findYearClosestToRef } from "../../../calculation/years";
+import { MONTH_DICTIONARY } from "../constants";
+import { ORDINAL_NUMBER_PATTERN, parseOrdinalNumberPattern } from "../constants";
+import { YEAR_PATTERN, parseYear } from "../constants";
+import { matchAnyPattern } from "../../../utils/pattern";
+import { AbstractParserWithWordBoundaryChecking } from "../../../common/parsers/AbstractParserWithWordBoundary";
+
+const PATTERN = new RegExp(
+    `(${matchAnyPattern(MONTH_DICTIONARY)})` +
+        "(?:-|/|\\s*,?\\s*)" +
+        `(${ORDINAL_NUMBER_PATTERN})(?!\\s*(?:am|pm))\\s*` +
+        "(?:" +
+        "(?:al|\\-|\\alle|\\del|\\s)\\s*" +
+        `(${ORDINAL_NUMBER_PATTERN})\\s*` +
+        ")?" +
+        "(?:" +
+        "(?:-|/|\\s*,?\\s*)" +
+        `(${YEAR_PATTERN})` +
+        ")?" +
+        "(?=\\W|$)(?!\\:\\d)",
+    "i"
+);
+
+const MONTH_NAME_GROUP = 1;
+const DATE_GROUP = 2;
+const DATE_TO_GROUP = 3;
+const YEAR_GROUP = 4;
+
+/**
+ * The parser for parsing US's date format that begin with month's name.
+ *  - January 13
+ *  - January 13, 2012
+ *  - January 13 - 15, 2012
+ * Note: Watch out for:
+ *  - January 12:00
+ *  - January 12.44
+ *  - January 1222344
+ */
+export default class ENMonthNameMiddleEndianParser extends AbstractParserWithWordBoundaryChecking {
+    innerPattern(): RegExp {
+        return PATTERN;
+    }
+
+    innerExtract(context: ParsingContext, match: RegExpMatchArray) {
+        const month = MONTH_DICTIONARY[match[MONTH_NAME_GROUP].toLowerCase()];
+        const day = parseOrdinalNumberPattern(match[DATE_GROUP]);
+        if (day > 31) {
+            return null;
+        }
+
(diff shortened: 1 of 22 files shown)


federicodilo:
I'm still getting some issues from tests, I guess I'll take some hints if possible 🙃

src/locales/it/parsers/ITMonthNameMiddleEndianParser.ts
```diff
@@ -0,0 +1,87 @@
...
+ *  - 10-25 marzo 2019
+ *  - 1° ago 2019
+ *  - 1° settembre 200 a.C.
+ *  - 1° settembre 2002 d.C.
+ *  - 19 gennaio 87
+ *  - 12 giugno 2013
+ *  - 1° novembre 2013
+ */
+
+export default class ITMonthNameMiddleEndianParser extends AbstractParserWithWordBoundaryChecking {
```
wanasit, line 49: From your example and test, this is Little-endian (i.e., day-month-year), not Middle-endian.

test/it/it.test.ts
```diff
@@ -0,0 +1,181 @@
...
+test("Test - Date + Time Expression", function () {
+    testSingleCase(chrono, "Qualcosa accade il 18/04/2014 13:00-16:00 appena", (result) => {
+        expect(result.text).toBe("18/04/2014 13:00-16:00");
+
+        expect(result.start).toBeDate(new Date(2014, 4 - 1, 18, 13));
+        expect(result.end).toBeDate(new Date(2014, 4 - 1, 18, 16));
+    });
+});
+
+test("Test - Time Expression", function () {
```
wanasit, line 15: I would move this to `it_time_exp.test.ts`

test/it/it.test.ts
```diff
@@ -0,0 +1,181 @@
...
+});
+
+test("Test - Random text", function () {
+    testSingleCase(chrono, "Adam <Adam@supercalendar.com> написал(а):\nIl giorno è 02.07.2013", (result) => {
+        expect(result.text).toBe("02.07.2013");
+    });
+
+    testSingleCase(chrono, "174 1 Novembre 2001-31 Marzo 2002", (result) => {
+        expect(result.text).toBe("1 Novembre 2001-31 Marzo 2002");
+    });
```
wanasit, line 61: I would be nice if you could also add `it_month_name_little_endian.test.ts` or just `it_month_name.test.ts` for testing pattern with month name.


In addition to these "DD MMMM YYYY" and "DD MMMM YYYY ~ DD MMMM YYYY", I think we have not tested, the "DD - DD, MMMM YYYY" pattern yet.

test/it/it_time_units_ago.test.ts
```diff
@@ -0,0 +1,260 @@
+import * as chrono from "../../src/";
+import { testSingleCase, testUnexpectedResult } from "../test_util";
+
+test("Test - Single Expression", function () {
```
wanasit, line 4: Could we also have `it_time_units_later.test.ts`? 
or we could rename this `t_time_units.test.ts` and add the time later test cases here?

wanasit (changes requested):
Hello, @federicodilo. Thank you very much from the change!! 

I added some comment. They are mostly around where could we test more.

I also recommend that you try running the test with coverage and see where we missing the test coverage (e.g. if you use WebStorm, [it should be just one-click](https://www.jetbrains.com/help/webstorm/code-coverage.html). other IDE should have similar functionality). We don't need to get it to 100% or 90%, but it good to know we have cover most functionality that we want to support.

federicodilo:
Well, I fixed some translations and tried to run the tests of which, unfortunately, some I could not get to work properly.

There are still some rather serious critical issues arose in dealing with the ambiguity of some sentences.

Eg.: There's an ambiguity in `ITTimeExpressionParser.ts` generated by "alle" in the followingPhase, its presence is however mandatory in both primaryPrefix and followingPhase ("_dalle hh:mm alle hh:mm_" meaning "_from hh:mm to hh:mm_" but using just "_alle hh:mm_" would mean "_at hh:mm_").

Morover, as before, "_all'_" can be used both for literal time units and literal days of the month.

Unfortunately, ambiguities are a constant in more or less all languages and because of my knowledge of JS, I wouldn't know how to come up with it. 😕 

Another problem that kept me busy was handling the digits at the beginning of the sentence as in `it_time_units_ago.test.ts`. I've really no clues on what's not working.

Still needing an helping hand if possible and still willing to contribute. 🙂

Thanks,
Federico

wanasit:
Hello @federicodilo. 
Yes. The problem sounds difficult. I'm really sorry that I cannot really do much to help. 

I think the problem is **we try to support too many patterns/cases from the beginning**. We should not try to make it as good as English for the first IT support. We should start recognizing a few Italian issues and add more patterns (+ tests). Then, we add more patterns and debug over time. That is how I was able to handle the complexity in other languages.
 
I know both you and @ntotao work very hard on translating everything, but **I think it would be better for us if you could revert/re-create a smaller/minimal and easier-to-test MR**. 

Here are what I think we leave out for future changes:
* Not supporting the time range yet (e.g. only "alle hh:mm", not "dalle hh:mm alle hh:mm")
* Not supporting the time units yet (both ago/later)
* (If it really difficult) Not support time expression yet (only date)
* Or just set up the IT configuration with one or two extractors you are confident (+tests, of course) would be enough for the first IT support MR.  

FYI. If you are participating [Hacktoberfest](https://hacktoberfest.com/), breaking things into multiple MRs is also better for you. If you send smaller MRs, I'll try to review and merge them as quickly as possible.

PR #476, opened 2022-10-24
docs: Corrects supported locale options

mikemikhaylov:
Adds a little redundancy to README, since the same list is specified on line 40:

> The current fully supported languages are `en`, `ja`, `fr`, `nl` and `ru` (`de`, `pt`, and `zh.hant` are partially supported). 

But maybe it's more clear to have the full list also in **Locale** section.


wanasit:
Thanks!

PR #482, opened 2022-11-30
More options for ru "after"

mikemikhaylov:
Just another option for RUTimeUnitCasualRelativeFormatParser


wanasit:
Thanks for the improvement.

PR #485, opened 2022-12-10
Basic spanish support

S4M3R:
Cloned and translated from Portuguese implementation due is very similar to Spanish, probably will be making more improvement on the future, based on feed back received by more Spanish-speaking users.


wanasit:
Hey @S4M3R. Thanks for the improvement!

Not trying to support all features is good. I'd actually prefer even multiple smaller changes.

PR #486, opened 2022-12-18
☀️ Support "midday" casual time in English

kieranm:
Small change to support "midday" as well as "noon"


wanasit:
@kieranm Merry Christmas. Thanks for the fix.

It'd be nice if you could also add some tests for the new pattern next time.

PR #489, opened 2022-12-30
Add until/till/through as date/time range expression operators

andrewpomeroy:
Until/till/through were being used as date/time-range joining operators in certain parsers, but not all. I found two parsers in which it made sense to support these operators, and added them in.


wanasit:
Hello. Sorry for the delay and thanks for the improvement!

andrewpomeroy:
No problem! Thanks so much for your great work on this library.

PR #495, opened 2023-01-27
[ES] added TimeUnitWithin parser for Spanish

trothe:
Ported the time unit within parser and tests from French to Spanish.


wanasit:
@trothe Thanks for the improvement!

PR #496, opened 2023-01-27
[DE] added TimeUnitWithin parser for German

trothe:
Added Time Unit Within parser and tests for German. I am a native German speaker.


wanasit:
Thanks for the improvement here too

PR #497, opened 2023-02-02
Add 'night' to casual reference

Slapbox:
Defaulting to 10pm. Initially I intended to make it simply an alias of `night`, but since `lastNight` defaults to midnight it seems wrong.

### DIFF
--- a/src/common/casualReferences.ts
+++ b/src/common/casualReferences.ts
@@ -82,6 +82,13 @@ export function evening(reference: ReferenceWithTimezone, implyHour = 20): Parsi
     return component;
 }
 
+export function night(reference: ReferenceWithTimezone, implyHour = 22): ParsingComponents {
+    const component = new ParsingComponents(reference, {});
+    component.imply("meridiem", Meridiem.PM);
+    component.imply("hour", implyHour);
+    return component;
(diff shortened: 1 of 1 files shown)


Slapbox:
@wanasit any thoughts on this one?

wanasit:
To rephrase my previous comment, even if we merge the change, the added `night()` function would be a dead code. 

Could you also include the change to use the `night()` function? 
Or wait until we have a chance to utilize it?

wanasit:
Ops. I just realized I didn't click to send the review.

Slapbox:
It turns out I think this is unnecessary. I had tested with "night" and it seemed not to work, but that must have been a consquence of some custom refiner on my end or something because as I went through the code I see that `night` is aliased to `evening` here: 

https://github.com/wanasit/chrono/blob/c1c782498694d92cb8ce42602830c3cadf356721/src/locales/en/parsers/ENCasualTimeParser.ts#L16

I think that's fine and doesn't need to be split further between `evening` and `night`. Adding support for a separate "night" would be a "nice-to-have" but since it works well enough for our use cases, I'll close this.

Thanks very much @wanasit.

PR #500, opened 2023-02-16
Add support for 'mrt' Dutch shortened month

skoppers:
March in Dutch is usually shortened as 'mrt', however currently only 'mar' is supported. This adds support for 'mrt'.


wanasit:
Hello. I don't speak Dutch, so I am going to assume that you have tested it.
Thanks for the improvement!

PR #509, opened 2023-03-23
Expose individual locales

jdicarlo-dh:
Hello! This PR is to allow importing individual locales from Chrono without all of them being imported implicitly.

### What changed

* `package.json` now exposes individual locales, which can be imported like this:
```js
// ESM
// TS Note: `moduleResolution` must be set to `node16` or `nodeNext` in tsconfig.json`
import * as chrono from 'chrono-node/en'
// CJS
const chrono = require('chrono-node/en')
```
* Interfaces defined in the root `index.ts` now live in their own file called `parsing.ts`
* Locale `index.ts` files now expose the types in `parsing.ts` and no longer depend on the root `index.ts`
* Non-English locales no longer implicitly depend on the `en` locale

### Why?

We're using Chrono with tools that utilize a single static binary for Node.js. These binaries are built with the `--without-intl` flag to reduce their size, which is fine for our use case as we don't need to support languages besides English.

However, when using the chrono library, other locales such as `ru` are loaded implicitly which causes an error due to the intl module not being present:
```
Invalid regular expression: /* omitted */: Invalid property name in character class
```

### Other Notes

These changes do not break the library's API and none of the tests required any modification. There were 5 failures when I ran tests locally, but I also got these errors when running tests on the master branch, so I don't think they're related to my changes. I have also updated the `README.md` with information about using the individual locales

Please let me know if I've missed anything or if there are any other questions/comments/concerns, and thank you for such a useful library! 😄 ❤️


wanasit:
Hello @jdicarlo-dh. Thanks for working on this. 

I agree with your idea, but because your change is big, I'll need to take some time to review and try it. It'd be nice if you could resolve the conflict with `src/index.ts` too.

jdicarlo-dh:
No worries, take your time! I will gladly fix up the conflict in the meantime

jdicarlo-dh:
@wanasit I rebased my branch with master and made some changes to integrate those changes appropriately with mine. I also updated the exports on the locale indexes to match what's exported by the root index. 

@JoakimNil would love to get your eyes on this as well!

jdicarlo-dh:
Hi @wanasit just wondering if you've had a chance to take a look at this yet. I've rebased it with the most recent changes from `master` so it should be good to go for testing 😄

src/parsing.ts
wanasit, line 1: Can you rename this to `types.ts` or `common.ts`?

README.md
```diff
@@ -214,6 +214,28 @@ chrono.ja.parseDate('昭和６４年１月７日');
...
 Current supported locale options are: `en`, `ja`, `fr`, `nl` and `ru` (`de`, `pt`, and `zh.hant` are partially supported).
 
+#### Importing specific locales
+
+Chrono exports all locale options by default for simplicity. However, this can cause issues when using Chrono if you're using a Node.js runtime that was built with the Intl module disabled (with the `--without-intl` flag), such as: 
+```
+Invalid regular expression: /* omitted */: Invalid property name in character class
+```
+
+This is because the Intl module is required to handle special characters, such as Cyrillic (`ru`).
```
wanasit, line 224: Do you know exactly which regex or file that cause this? Could you file a separate issue for this?

package.json
```diff
@@ -26,6 +26,14 @@
     ".": {
       "require": "./dist/cjs/index.js",
       "import": "./dist/esm/index.js"
+    },
+    "./*": {
+      "require": "./dist/cjs/locales/*/index.js",
+      "import": "./dist/esm/locales/*/index.js"
+    },
+    "./*/*": {
```
wanasit, line 34: What is this for? I don't think we have index.ts in these sub-directories. 

If it's necessary, lets add that later.

wanasit:
Hello @jdicarlo-dh. 

Could you look at my comments and make some naming changes? Overall, the change looks good!

Also, if you know which RegEx in `ru` cause the problem, could you open a separate issue for that?

README.md
jdicarlo-dh: None of the regexes are actually invalid, the issue comes from instances of using Chrono with a build of Node.js that doesn't include the `Intl` module. Most static binaries, such as what `vercel/pkg` and `caoccao/Javet` use, disable the module to reduce binary size. Without the module enabled, Node.js _thinks_ that the regex is invalid because it doesn't know how to handle special characters such as Cyrillic. I don't think there is any way to "fix" this because it's not really broken, it's just the reality of using a binary without `Intl` enabled.

package.json
jdicarlo-dh: I had to add these to be able to import the individual locales without importing the entire library. Each locale has an `index.ts` that builds to `/dist/cjs/<locale>/index.js` and `/dist/esm/<locale>/index.js` respectively:

image

So by adding these it allows all locales to be imported individually either as CJS or ESM: 
```jsonc
// import all of chrono, like normal 
".": {
  // const chrono = require('chrono-node')
  "require": "./dist/cjs/index.js",
  // import chrono from 'chrono-node'
  "import": "./dist/esm/index.js"
},
// import only one locale (`en`)
"./*": {
  // const chrono = require('chrono-node/en')
  "require": "./dist/cjs/locales/*/index.js",
  // import chrono from 'chrono-node/en'
  "import": "./dist/esm/locales/*/index.js"
},
// import only one locale, with subtype (`zh/hant`)
"./*/*": {
  // const chrono = require('chrono-node/zh/hant')
  "require": "./dist/cjs/locales/*/*/index.js",
  // import chrono from 'chrono-node/zh/hant'
  "import": "./dist/esm/locales/*/*/index.js"
}
```

src/parsing.ts
jdicarlo-dh: done!

jdicarlo-dh:
@wanasit I changed the name from `parsing.ts` to `types.ts` and replied to your other comments 😄

package.json
wanasit: I see. I forgot about `/zh/hant`.

PR #510, opened 2023-03-28
Adds more options for minutes in ru locale

mikemikhaylov:
- Adds more forms of "minute" in ru locale 
- Changes `let` to `const` where let is redundant


wanasit:
Thanks for the changes. 
(I don't speak Russia, so I'll just assume that this is correct.)

PR #511, opened 2023-03-30
[JA] Add support for Hiragana

yuichkun:
# TL;DR

This PR adds support for Hiragana(ひらがな) in the Japanese date parser.
I didn't know the best way of treating Hiragana/Kanji, so for now I just did it in the most straight forward, or redundant way.
If you have any thoughts on how to improve this, I'd love to challenge myself!

In addition to adding support for Hiragana, I also added some test cases that seemed to be missing.

Btw, thank you so much for writing this library, and I really appreciate how well this library is maintained.

### DIFF
--- a/src/locales/ja/parsers/JPCasualDateParser.ts
+++ b/src/locales/ja/parsers/JPCasualDateParser.ts
@@ -3,7 +3,30 @@ import dayjs from "dayjs";
 import { Meridiem } from "../../../index";
 import * as references from "../../../common/casualReferences";
 
-const PATTERN = /今日|当日|昨日|明日|今夜|今夕|今晩|今朝/i;
+const PATTERN = /今日|きょう|当日|とうじつ|昨日|きのう|明日|あした|今夜|こんや|今夕|こんゆう|今晩|こんばん|今朝|けさ/i;
+
+function normalizeTextToKanji(str: string) {
+    switch (str) {
+        case "きょう":
+            return "今日";
+        case "とうじつ":
+            return "当日";
+        case "きのう":
+            return "昨日";
+        case "あした":
+            return "明日";
+        case "こんや":
+            return "今夜";
+        case "こんゆう":
+            return "今夕";
+        case "こんばん":
+            return "今晩";
+        case "けさ":
+            return "今朝";
+        default:
+            return str;
+    }
+}
 
 export default class JPCasualDateParser implements Parser {
     pattern() {
@@ -11,7 +34,7 @@ export default class JPCasualDateParser implements Parser {
     }
 
     extract(context: ParsingContext, match: RegExpMatchArray) {
-        const text = match[0];
+        const text = normalizeTextToKanji(match[0]);
 
         const date = dayjs(context.refDate);
         const components = context.createParsingComponents();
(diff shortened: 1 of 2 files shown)


test/ja/ja_casual.test.ts
```diff
@@ -2,27 +2,90 @@ import * as chrono from "../../src/";
 import { testSingleCase } from "../test_util";
 
 test("Test - Single Expression", function () {
-    testSingleCase(chrono.ja, "今日感じたことを忘れずに", new Date(2012, 8 - 1, 10, 12), (result) => {
-        expect(result.index).toBe(0);
-        expect(result.text).toBe("今日");
+    function testToday(pattern: string) {
```
wanasit, line 5: Could you not use a function like this and repeat testSingleCase() for each case?

I know it's a lot of duplicated code, but we want to ensure the tests are straightforward. (e.g. if the expect fails we will know which line and case)

test/ja/ja_casual.test.ts
```diff
@@ -2,27 +2,90 @@ import * as chrono from "../../src/";
...
+        testSingleCase(chrono.ja, `${pattern}感じたことを忘れずに`, new Date(2012, 8 - 1, 10, 12), (result) => {
+            expect(result.index).toBe(0);
+            expect(result.text).toBe(pattern);
+            expect(result.start).not.toBeNull();
+            expect(result.start.get("year")).toBe(2012);
+            expect(result.start.get("month")).toBe(8);
+            expect(result.start.get("day")).toBe(10);
+            expect(result.start).toBeDate(new Date(2012, 8 - 1, 10, 12));
```
wanasit, line 13: I think, previously, we test too many things. e.g. only `expect(result.start).toBeDate()` and maybe `expect(result.text)` is enough.

src/locales/ja/parsers/JPCasualDateParser.ts
```diff
@@ -3,15 +3,38 @@ import dayjs from "dayjs";
 import { Meridiem } from "../../../index";
 import * as references from "../../../common/casualReferences";
 
-const PATTERN = /今日|当日|昨日|明日|今夜|今夕|今晩|今朝/i;
+const PATTERN = /今日|きょう|当日|とうじつ|昨日|きのう|明日|あした|今夜|こんや|今夕|こんゆう|今晩|こんばん|今朝|けさ/i;
+
+function normalizeTextToKanji(str: string) {
```
wanasit, line 8: Nit: `text` not `str`, because your function name is normalizeTextToKanji, not normalizeStrToKanji.

wanasit:
Hello @yuichkun. Thank you so much for the change.

Could you please take a look at my comments, especially about the test?
It would be nice if you can address them.

src/locales/ja/parsers/JPCasualDateParser.ts
yuichkun: Agree!

test/ja/ja_casual.test.ts
yuichkun: Will do so accordingly 👍

test/ja/ja_casual.test.ts
yuichkun: > we want to ensure the tests are really straightforward

I absolutely agree with you. 
I'll make the tests like so and maybe with less assertions like you mentioned [here](https://github.com/wanasit/chrono/pull/511#discussion_r1161188815)

yuichkun:
Hi @wanasit ,
thank you for your feedback, and I think I've fixed the points you mentioned.
Please let me know if there's anything else I should reconsider!

PR #520, opened 2023-07-18
Export `ParsedComponents` again

SferaDev:
We are using the public `ParsedComponents` type, however in https://github.com/wanasit/chrono/commit/feb250d3ca44ff30efb59d62440e576496434d8f when the interfaces where moved `parsing.ts` and later to `types.ts`, `ParsedComponents` was not re-exported again.

There are other interfaces that went private with that change, up to you to decide which ones you want to re-export.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -2,7 +2,7 @@ import { DebugHandler, DebugConsume } from "./debugging";
 import * as en from "./locales/en";
 import { Chrono, Parser, Refiner } from "./chrono";
 import { ParsingResult } from "./results";
-import { Component, ParsedResult, ParsingOption, ParsingReference, Meridiem, Weekday } from "./types";
+import { Component, ParsedComponents, ParsedResult, ParsingOption, ParsingReference, Meridiem, Weekday } from "./types";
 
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -2,7 +2,7 @@ import { DebugHandler, DebugConsume } from "./debugging";
 import * as en from "./locales/en";
 import { Chrono, Parser, Refiner } from "./chrono";
 import { ParsingResult } from "./results";
-import { Component, ParsedResult, ParsingOption, ParsingReference, Meridiem, Weekday } from "./types";
+import { Component, ParsedComponents, ParsedResult, ParsingOption, ParsingReference, Meridiem, Weekday } from "./types";
 
 export { en, Chrono, Parser, Refiner, ParsingResult };
```
zhouzi, line 7: `ParsedComponents` needs to be exported here.

SferaDev: Done! Sorry, that's what I get for not cloning and opening it in the IDE... 🤦🏻

wanasit:
Thanks both of you.

PR #525, opened 2023-08-10
Add casualText to expose casual match

### LINKED ISSUE #524: Expose casual date match
I'm building an interface where the user can type "tmrw" and the autocomplete suggests "Tomorrow".

So I need to know which casual date matched.

There's already a switch in the casual parsers discrimating the different casual dates so I think it can be added with no breaking changes.

I'm looking into it and will suggest a PR.

zhouzi:
This PR adds casualText to ParsedComponents for user to identify which casual parser matched.

For now I only added it in a few places but I'm happy to add it everywhere if you think it makes sense.

Resolve #524

### DIFF
--- a/src/chrono.ts
+++ b/src/chrono.ts
@@ -159,12 +159,15 @@ export class ParsingContext implements DebugHandler {
         this.refDate = this.reference.instant;
     }
 
-    createParsingComponents(components?: { [c in Component]?: number } | ParsingComponents): ParsingComponents {
+    createParsingComponents(
+        components?: { [c in Component]?: number } | ParsingComponents,
+        casualText?: string
+    ): ParsingComponents {
         if (components instanceof ParsingComponents) {
             return components;
         }
 
-        return new ParsingComponents(this.reference, components);
+        return new ParsingComponents(this.reference, components, casualText);
     }
 
     createParsingResult(
--- a/src/common/casualReferences.ts
+++ b/src/common/casualReferences.ts
@@ -9,9 +9,9 @@ import {
 } from "../utils/dayjs";
 import { Meridiem } from "../types";
 
-export function now(reference: ReferenceWithTimezone): ParsingComponents {
+export function now(reference: ReferenceWithTimezone, casualText?: string): ParsingComponents {
     const targetDate = dayjs(reference.instant);
-    const component = new ParsingComponents(reference, {});
+    const component = new ParsingComponents(reference, {}, casualText);
     assignSimilarDate(component, targetDate);
     assignSimilarTime(component, targetDate);
     if (reference.timezoneOffset !== null) {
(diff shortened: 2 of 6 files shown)


wanasit:
Hello @zhouzi. Thanks for reaching out. I'm sorry for not getting back to you sooner.

Adding a new feature to the interface is a significant change for all users and contributors. Unfortunately, I am unable to accept this change in its current form.

For your use case, I suggest finding a way to implement this outside of Chrono. 
I also think keeping the date parsing and formatting/suggesting problems separated may be more beneficial for you.  

If you can provide more information about your app (e.g. how it should handle inputs like "tmr 4:20 am" or "September 2nd"), I may be able to suggest additional ideas.

--

If we were to add this feature, we should aim for the feature be more generalized and work for different cases and locales. This would take significantly more work. For example:
```typescript
export interface ParsedComponents {
  ...
  /**
   * Get the normalized or formatted text for explaining this date component.
   * - "tmr 4:20 am" -> "Tomorrow at 4:20AM"
   * - "Aug 12, 2023" -> "2023-08-12" or "August 12, 2023"
   * - "12th Aug 2023" -> "2023-08-12" or "August 12, 2023"
   * - "12min later" -> "12 minutes later"
   * - "+2m" -> "2 minutes later"
   */
  normalizedText(locale?): string;
}
```

zhouzi:
> Adding a new feature to the interface is a significant change for all users and contributors. Unfortunately, I am unable to accept this change in its current form.

Agreed, to be honest I made the least amount of changes possible so it works for our own use case. I thought we'd probably need to have a discussion before going further.

> For your use case, I suggest finding a way to implement this outside of Chrono. I also think keeping the date parsing and formatting/suggesting problems separated may be more beneficial for you.

The thing is we like and need everything from Chrono. But I'll share more about the use case so perhaps you have a suggestion.

> If you can provide more information about your app

We have an app with a list of notes.

Users can create filters to only show notes that have been created after a given date.

When creating the filter they can use natural language to input the date.

So when they type "yesterday" it can mean two things:

- Either they want to only list the notes that have been created after "yesterday", yesterday being a dynamic date (the date before today)
- Or they want to only list the notes that have been created after the 2023-08-19 (provided today is the 2023-08-20)

So what we do when they type "yesterday" is that we suggest two results:

- August 19, 2023 (fixed date)
- Yesterday (dynamic date)

If they select the fixed date, we save 2023-08-19 in the database. Otherwise we'd save "Yesterday".

And we use Chrono to parse the date when applying the filters.

> how it should handle inputs like "tmr 4:20 am" or "September 2nd"

In this scenario it's ok for us if it doesn't return a "casual text". But yes, I imagine the best result would be "Tomorrow 4:20am or "September 2nd". The idea is to return something that's nice to read and that Chrono can parse back.

---

I agree that it might not make a lot of sense to have it in the core library if it's not a common use case.

Let me know if you have any suggestions!

In the mean time I'll think more about it and see if I can come up with something that would include minor changes, if any.

zhouzi:
This PR has been superseded by #534 

I will update the original issue #524 with an example of how we are solving our use case with the tags and/or send a PR adding documentation.

PR #530, opened 2023-08-21
update regex refine french, add 'au'

maxwarch:
I add another article _"au"_ for french version, to split start and end dates in the right place.


wanasit:
Hello. Thanks for the improvement.

PR #534, opened 2023-09-02
New: Introduce Parsing Tags

wanasit:
# Parsing Tags

A parsing tag or just "tag" is a string attached to parsed components (and results) by parsers and refiners for additional information about the location and context where components were created or modified.

```typescript
const custom = chrono.casual.clone();
custom.parsers.push({
    pattern: () => { return /\bChristmas\b/i },
    extract: (context, match) => {
        return .createParsingComponents({day: 25, month: 12 })
                .addTag("USHoliday/chirstmas");
    }
});
custom.parsers.push({
    pattern: () => { return /\bNew Year\b/i },
    extract: (context, match) => {
        return .createParsingComponents({day: 1, month: 1 })
                .addTag("USHoliday/new_year");
    }
});

....

const result = custom.parse("I'll arrive at 2.30AM on Christmas night....")[0];

// To recognize the result from a specific parser... 
result.tags().has("CustomHoliday/chirstmas")

// To recognize the result from a specific parser group... 
Array.from(result.tags()).some((t) => t.startsWith("USHoliday/")
Array.from(result.tags()).some((t) => t.match(/EN\w+Parser/)

// To recognize the special content/info from paser
const theHolidayName = Array.from(result.tags())
    .find(t => t.startsWith("USHoliday/"))
   ?.replace("USHoliday/", "");
```

### Debugging Extraction: 

Inside Chrono, multiple parsers and refiners work together to produce the result. Identifying which component(s) output the incorrect results is challenging.  

By tagging the result with involved parsers and refiners, it should be easier to identify the area to fix from the bug report or debug log:
```text
Error: Expected date to be: Sat Sep 02 2023 14:29:21 GMT+0900 (Japan Standard Time) Received: Sun Sep 03 2023 14:29:21 GMT+0900 (Japan Standard Time) ([ParsingComponents {
            tags: ["ENCasualDateParser/extract/tomorrow"], 
            knownValues: {"day":3,"month":9,"year":2023}, 
            impliedValues: {"hour":14,"minute":29,"second":21,"millisecond":371}}, 
            reference: {"instant":"2023-09-02T05:29:21.371Z"}])
```

### Recognizing results "types" (or "categories")

These is the list of example ideas:
* Check which parser's method produce the result and treat the output accordingly (e.g. format the output text in #524)
* Check if the output is in which locale (e.g. from "EN-" or common ISO parser)

### DIFF
--- a/src/locales/en/parsers/ENCasualDateParser.ts
+++ b/src/locales/en/parsers/ENCasualDateParser.ts
@@ -15,25 +15,35 @@ export default class ENCasualDateParser extends AbstractParserWithWordBoundaryCh
     innerExtract(context: ParsingContext, match: RegExpMatchArray): ParsingComponents | ParsingResult {
         let targetDate = dayjs(context.refDate);
         const lowerText = match[0].toLowerCase();
-        const component = context.createParsingComponents();
+        let component = context.createParsingComponents();
 
         switch (lowerText) {
             case "now":
-                return references.now(context.reference);
+                component = references.now(context.reference);
+                component.addTag("ENCasualDateParser/extract/now");
+                break;
 
             case "today":
-                return references.today(context.reference);
+                component = references.today(context.reference);
+                component.addTag("ENCasualDateParser/extract/today");
+                break;
 
             case "yesterday":
-                return references.yesterday(context.reference);
+                component = references.yesterday(context.reference);
+                component.addTag("ENCasualDateParser/extract/yesterday");
+                break;
 
             case "tomorrow":
             case "tmr":
             case "tmrw":
-                return references.tomorrow(context.reference);
+                component = references.tomorrow(context.reference);
+                component.addTag("ENCasualDateParser/extract/tomorrow");
+                break;
 
             case "tonight":
-                return references.tonight(context.reference);
+                component = references.tonight(context.reference);
+                component.addTag("ENCasualDateParser/extract/tonight");
+                break;
 
             default:
                 if (lowerText.match(/last\s*night/)) {
@@ -43,6 +53,7 @@ export default class ENCasualDateParser extends AbstractParserWithWordBoundaryCh
 
                     assignSimilarDate(component, targetDate);
                     component.imply("hour", 0);
+                    component.addTag("ENCasualDateParser/extract/last_night");
                 }
                 break;
         }
--- a/src/locales/en/parsers/ENCasualDateParser.ts
+++ b/src/locales/en/parsers/ENCasualDateParser.ts
@@ -20,29 +20,24 @@ export default class ENCasualDateParser extends AbstractParserWithWordBoundaryCh
         switch (lowerText) {
             case "now":
                 component = references.now(context.reference);
-                component.addTag("ENCasualDateParser/extract/now");
                 break;
 
             case "today":
                 component = references.today(context.reference);
-                component.addTag("ENCasualDateParser/extract/today");
                 break;
 
             case "yesterday":
                 component = references.yesterday(context.reference);
-                component.addTag("ENCasualDateParser/extract/yesterday");
                 break;
 
             case "tomorrow":
             case "tmr":
             case "tmrw":
                 component = references.tomorrow(context.reference);
-                component.addTag("ENCasualDateParser/extract/tomorrow");
                 break;
 
             case "tonight":
                 component = references.tonight(context.reference);
-                component.addTag("ENCasualDateParser/extract/tonight");
                 break;
 
             default:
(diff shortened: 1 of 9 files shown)


src/locales/en/parsers/ENCasualDateParser.ts
```diff
@@ -15,25 +15,35 @@ export default class ENCasualDateParser extends AbstractParserWithWordBoundaryCh
...
             case "today":
-                return references.today(context.reference);
+                component = references.today(context.reference);
+                component.addTag("ENCasualDateParser/extract/today");
+                break;
 
             case "yesterday":
-                return references.yesterday(context.reference);
+                component = references.yesterday(context.reference);
+                component.addTag("ENCasualDateParser/extract/yesterday");
```
zhouzi, line 33: I think we need a generic tag for that use case as well: `casualReference/yesterday`. This way users can identify the casual reference regardless of the language (or having to loop over all of them).

Same for the other matches below: `casualReference/tomorrow`, `casualReference/tonight`, `casualReference/last_night`.

Perhaps we can add another tag to identify the parser: `parser/ENCasualDateParser`.

zhouzi (commented):
I like this idea a lot 👍

One limitation is that the parsers don't provide the translated casual reference.

Translation would need to be done in user-land.

Perhaps that's no big deal if the library provides utilities for that. For example (pseudo code):

```javascript
export function translateCasualReference(component) {
  const casualReference = component.getTag('casualReference/');
  
  if (casualReference == null) {
    return undefined;
  }
  
  const parser = getParser(component.getTag('parser/'));
  return parser.messages[casualReference];
}
```

Another option would be to turn tags into an object instead of an array. This way we could have:

```
{
  "casualReference": {
    "kind": "today" | "tonight" | ...,
    "value": "Today" | "Aujourd'hui" | "Hoy"
  }
}
```

src/locales/en/parsers/ENCasualDateParser.ts
wanasit: Thanks for the feedback. I'm also considering if we should add the tag in `ENCasualDateParser` or the `casualReferences`.

Seem that concepts such as 'tomorrow' should be tagged by 'casualReference', so it works orthogonally to the locale-specific. 

Maybe we could also simply add in both places. I don't think adding tags is that costly.

wanasit:
> One limitation is that the parsers don't provide the translated casual reference. 
> Translation would need to be done in user-land.

Yes. I'm afraid that is the case. 

Providing translation or date formatting (back into localized human-readable) is not want I am willing to support inside Chrono for now. I hope there are other libraries that are more specialized in doing that, or you have to add that on your application side.

That said, after you try adding the translation to your application, if you still think it is more fitting to support that feature inside Chrono. I'm open to the discussion.

> Another option would be to turn tags into an object instead of an array. 

I am also considering having a value associated with the tag (aka. defining `tags` as `Map`, not `Set`), but that feels it could get over-complicated. If different tags from different parsers/refiners can have different values, that would be difficult to keep track of.

----

Let me try a few more ideas and fix the issues.
I plan to release the feature next week.

src/locales/en/parsers/ENCasualDateParser.ts
zhouzi: > Seem that concepts such as 'tomorrow' should be tagged by 'casualReference', so it works orthogonally to the locale-specific.

Agreed.

> Maybe we could also simply add in both places. I don't think adding tags is that costly.

I thought about that too, adding `ENCasualDateParser/yesterday` as well as `casualReference/yesterday`.

We would still need `parser/ENCasualDateParser` (or similar), though.

It seems to me that `casualReference/yesterday` and `parser/ENCasualDateParser` are enough to know that it's coming from the `ENCasualDateParser` for the casual reference `yesterday`.

But as you said, having several tags is no big deal, even if they are a little redundant. Might actually make the developer experience better.

zhouzi:
Agreed.

Honestly I think tags as currently implemented are already enough (I think the naming inconsistencies discussed in the comments need to be adressed though).

src/locales/en/parsers/ENCasualDateParser.ts
wanasit: I ended up separately adding tags in casualReference (`casualReference/yesterday`, `casualReference/evening`, etc) and parsers (`parser/ENCasualDateParser`, `parser/ENCasualTimeParser` etc) as you suggested.

I'll add tags in other parsers and refiners later, but please feel free to send PR to add more tags if there are specific parsers or contexts you need to recognize.

zhouzi: Sorry for the delay, I was on holidays.

This is looking great, I'll be updating the library this week or the next. I'll send a PR if needed but I think we have everything we need as it stands.

Thanks again 🙏

Issue #538, opened 2023-11-13
Main branch failing tests, preventing me from making PR


HendrikJan:
Hi,

I'm trying to create a PR for a minor improvement for the Dutch language.

Unfortunately I cannot create a commit because Husky won't let me.
Husky won't let me because the test "Compare with native js" fails.

Should I wait for a fix for this test before creating a PR?

Kind regards,
Hendrik Jan

HendrikJan:
I've now created a PR using "--no-verify", but questing remains if I should wait for the test to pass.

wanasit:
Hello @HendrikJan. 

Could you share more specific details about the cases that are failing? or what error message do you get?

HendrikJan:
The error I get is this:
![image](https://github.com/wanasit/chrono/assets/2218920/6a0c4daa-0ffb-4e24-b5d5-332004749bab)

Since you already accepted my PR this failing test is not blocking me anymore.

PR #539, opened 2023-11-13
Improve dutch

HendrikJan:
Correct in Dutch is "komend uur", not "komende uur".  
And there are some other combinations that were not detected yet: "aankomende maand", "komend jaar" and some more.

In this PR both  
"komend" and "komende",  
"volgend" and "volgende",  
"vorig" and "vorige",
"aankomend" and "aankomende"  
are allowed.

Changed test for "volgende maand" into "aankomende maand", as there are already tests for "volgende week", "volgende dag", "volgend jaar" and no tests for "aankomende".

### DIFF
--- a/src/locales/nl/parsers/NLRelativeDateFormatParser.ts
+++ b/src/locales/nl/parsers/NLRelativeDateFormatParser.ts
@@ -6,13 +6,13 @@ import { AbstractParserWithWordBoundaryChecking } from "../../../common/parsers/
 import { matchAnyPattern } from "../../../utils/pattern";
 
 const PATTERN = new RegExp(
-    `(dit|deze|(aan)?komend|volgend|afgelopen|vorig)e?\\s*(${matchAnyPattern(TIME_UNIT_DICTIONARY)})(?=\\s*)` +
+    `(dit|deze|(?:aan)?komend|volgend|afgelopen|vorig)e?\\s*(${matchAnyPattern(TIME_UNIT_DICTIONARY)})(?=\\s*)` +
     "(?=\\W|$)",
     "i"
 );
 
 const MODIFIER_WORD_GROUP = 1;
-const RELATIVE_WORD_GROUP = 3;
+const RELATIVE_WORD_GROUP = 2;
 
 export default class NLRelativeDateFormatParser extends AbstractParserWithWordBoundaryChecking {
     innerPattern(): RegExp {
--- a/src/locales/nl/parsers/NLRelativeDateFormatParser.ts
+++ b/src/locales/nl/parsers/NLRelativeDateFormatParser.ts
@@ -6,8 +6,8 @@ import { AbstractParserWithWordBoundaryChecking } from "../../../common/parsers/
 import { matchAnyPattern } from "../../../utils/pattern";
 
 const PATTERN = new RegExp(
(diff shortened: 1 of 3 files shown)


HendrikJan:
The test will currently fail because the main branch I forked from has a failing test. See #538

src/locales/nl/parsers/NLRelativeDateFormatParser.ts
```diff
@@ -6,13 +6,13 @@ import { AbstractParserWithWordBoundaryChecking } from "../../../common/parsers/
 import { matchAnyPattern } from "../../../utils/pattern";
 
 const PATTERN = new RegExp(
-    `(dit|deze|komende|volgend|volgende|afgelopen|vorige)\\s*(${matchAnyPattern(TIME_UNIT_DICTIONARY)})(?=\\s*)` +
-        "(?=\\W|$)",
+    `(dit|deze|(aan)?komend|volgend|afgelopen|vorig)e?\\s*(${matchAnyPattern(TIME_UNIT_DICTIONARY)})(?=\\s*)` +
```
wanasit, line 9: I would use non-capturing group  `(?:aan)?` instead of `(aan)`.

test/nl/nl_relative.test.ts
```diff
@@ -71,7 +71,7 @@ test("Test - Past relative expressions", () => {
 
 test("Test - Future relative expressions", () => {
     // next hour
-    testSingleCase(chrono.nl, "komende uur", new Date(2016, 10 - 1, 1, 12), (result, text) => {
```
wanasit, line 74: Could you add more tests for the version without "-e" instead of changing the current ones?

src/locales/nl/parsers/NLTimeUnitCasualRelativeFormatParser.ts
```diff
@@ -4,18 +4,21 @@ import { ParsingComponents } from "../../../results";
 import { AbstractParserWithWordBoundaryChecking } from "../../../common/parsers/AbstractParserWithWordBoundary";
 import { reverseTimeUnits } from "../../../utils/timeunits";
 
-const PATTERN = new RegExp(`(deze|vorige|afgelopen|komende|over|\\+|-)\\s*(${TIME_UNITS_PATTERN})(?=\\W|$)`, "i");
+const PATTERN = new RegExp(`(dit|deze|vorig|afgelopen|(aan)?komend|over|\\+|-)e?\\s*(${TIME_UNITS_PATTERN})(?=\\W|$)`, "i");
```
wanasit, line 7: Same here. It is better to use a non-capturing group if you don't really intend to capture the "aan" part.

wanasit (commented):
Thanks for the improvement. 

I added a few comments. It would be nice if you could address them, but if you don't have time I will merge it next week.

src/locales/nl/parsers/NLTimeUnitCasualRelativeFormatParser.ts
HendrikJan: Sure. Done.

test/nl/nl_relative.test.ts
wanasit: I see. I think we should go with your suggestion.

PR #542, opened 2023-12-20
[EN] Fix for multiple time units within parser (e.g. set a timer for 5 minutes and 30 seconds)

trothe:
I ran into issues with phrases like "for 5 minutes AND 30 seconds". It seems the regex didn't capture the second time unit. I updated the regex to capture multiple time units in ENTimeUnitWithinFormatParser. I also added a test case. I hope this is acceptable. If not let me know.

### DIFF
--- a/test/en/en_time_units_within.test.ts
+++ b/test/en/en_time_units_within.test.ts
@@ -60,6 +60,13 @@ test("Test - The normal within expression", () => {
         expect(result.start).toBeDate(new Date(2012, 7, 10, 12, 19, 30));
     });
 
+    testSingleCase(chrono, "set a timer for 10 minutes and 10 seconds", new Date(2012, 7, 10, 12, 14), { forwardDate: true },  (result) => {
+        expect(result.index).toBe(12);
+        expect(result.text).toBe("for 10 minutes and 10 seconds");
+
+        expect(result.start).toBeDate(new Date(2012, 7, 10, 12, 24, 10));
+    });
+
     testSingleCase(chrono, "within 1 hour", new Date(2012, 7, 10, 12, 14), (result) => {
         expect(result.index).toBe(0);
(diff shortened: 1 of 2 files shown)


test/en/en_time_units_within.test.ts
```diff
@@ -53,6 +53,20 @@ test("Test - The normal within expression", () => {
...
     });
 
+    testSingleCase(chrono, "set a timer for 5 minutes and 30 seconds", new Date(2012, 7, 10, 12, 14), (result) => {
+        expect(result.index).toBe(12);
+        expect(result.text).toBe("for 5 minutes and 30 seconds");
+
+        expect(result.start).toBeDate(new Date(2012, 7, 10, 12, 19, 30));
+    });
+
+    testSingleCase(chrono, "set a timer for 10 minutes and 10 seconds", new Date(2012, 7, 10, 12, 14), { forwardDate: true },  (result) => {
```
wanasit, line 63: I would also test a combination of both comma + and e.g. "1 hour, 10 minutes, and 10 seconds".

wanasit (changes requested):
Hello @trothe. Thanks for your improvement!

However, I think adding the connection word in the parser pattern is not the best way to fix the problem (e.g. we also want to support other parsers e.g. "5 minutes AND 30 seconds ago"). Instead, we should fix the `TIME_UNIT_PATTERN` in `constants.ts`.

I would:
* Update `repeatedTimeunitPattern` to take a connector as a parameter (with the current pattern as default). 
* For English constant, we use the current pattern plus your new connector.

Sorry if my request is too difficult.

wanasit:
Hello @trothe! Thanks again for the improvement.
I incorporated your change in d1a71c9ed8fdfd25ea1128565cb45bdfa67c27ff.

PR #544, opened 2024-01-15
Fix parseDate return type not nullable

goodoldneon:
Fix `parseDate` returning `Date` instead of `Date | null` in its types.

Setting `"strictNullChecks": true` in `tsconfig.json` will prevent regressions, but that requires some refactoring to satisfy. There are a handful of places that:
- Expect `null` but get a value that could be `undefined`
- Expect `undefined` but get a value that could be `null`
- Expect a non-nullish type but get a value that could be nullish


wanasit:
Thanks for the fix!

Issue #582, opened 2025-02-27
Timezone is not accounted for in certain scenarios


zachsents:
I've found that with certain expressions, the timezone option isn't applied to the specified time.

```javascript
parseDate("9am", { timezone: "PST" })
// 2025-02-27T17:00:00.000Z -- correct
parseDate("tomorrow at 9am", { timezone: "PST" })
// 2025-02-28T17:00:00.000Z -- correct
parseDate("in 2 weeks at 9am", { timezone: "PST" })
// 2025-03-13T09:00:00.000Z -- incorrect
parseDate("2 weeks ago at 9am", { timezone: "PST" })
// 2025-02-13T09:00:00.000Z -- incorrect
parseDate("next friday at 9am", { timezone: "PST" })
// 2025-03-06T17:00:00.000Z -- correct
```

As you can see, the time is still parsed, just not in the correct timezone. 9am is interpreted as 9am UTC in the problematic scenarios. This was done from an environment in UTC.

wanasit:
This is more difficult than I thought. The [#583](https://github.com/wanasit/chrono/pull/583) fix breaks the test in other timezones.

wanasit:
Sorry for taking a long time.

This has been fixed together with #546 in 4033a27afda9bf12239709ba983e511eed530c32. Because the fix modifies significant parts of the existing calculation, the change is rolled out as a new minor version update (v2.8.0). 

Please update to the newer version and check the release notes for details.

PR #583, opened 2025-02-27
Fix: timezone option for relative dates should affect time

zachsents:
Addressing #582 

I'm not intimately familiar with all of the logic going on here, but tests pass, so I'm hoping you'll find this to be a suitable solution.

Also added 2 tests which address these cases. I added them in the main en locale test file; I'm not sure if there's a better spot for them.

### DIFF
--- a/src/results.ts
+++ b/src/results.ts
@@ -227,7 +227,7 @@ export class ParsingComponents implements ParsedComponents {
         } else {
             implySimilarTime(components, date);
             if (reference.timezoneOffset !== null) {
-                components.imply("timezoneOffset", -reference.instant.getTimezoneOffset());
+                components.imply("timezoneOffset", reference.timezoneOffset ?? -reference.instant.getTimezoneOffset());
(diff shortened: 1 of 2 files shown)


KodeurKubik:
+1 on this :)

src/results.ts
```diff
@@ -227,7 +227,7 @@ export class ParsingComponents implements ParsedComponents {
         } else {
             implySimilarTime(components, date);
             if (reference.timezoneOffset !== null) {
-                components.imply("timezoneOffset", -reference.instant.getTimezoneOffset());
+                components.imply("timezoneOffset", reference.timezoneOffset ?? -reference.instant.getTimezoneOffset());
```
wanasit, line 230: We don't need the fallback here. The `reference.timezoneOffset` should not be null, as it was checked above.

wanasit (approved):
Thanks for the fix. Only one small comment.

PR #585, opened 2025-03-12
Improve Japanese

koji520:
## Changes

- I removed "当日" on `JPCasualDateParser`. Because "当日" means that day or the day, not today.  
- I added "本日" as today on `JPCasualDateParser`
- I added "~" and "～" as patternBetween on `JPMergeDateRangeRefiner`. They are often used in Japanese.


wanasit:
Thank you for the improvement!

PR #593, opened 2025-04-24
Fix noon to make hour a known value rather than an implied value

crazyjncsu:
Making "noon" consistent with "midnight".

This is important to me because (for some unknown reason) 12 is by default implied for hour.  So I have code to check to see if hour is "implied" vs "known", and I throw it out if it's implied and set it to zero.  But for this "noon" I was throwing it out, which is not desired.

An alternative solution for me would be to allow a default hour to be set in config rather than implying 12.

Should the minute/second/milliseconds also be known for both midnight and noon?  Doesn't a

### DIFF
--- a/src/common/casualReferences.ts
+++ b/src/common/casualReferences.ts
@@ -137,7 +137,7 @@ export function afternoon(reference: ReferenceWithTimezone, implyHour = 15): Par
 export function noon(reference: ReferenceWithTimezone): ParsingComponents {
     const component = new ParsingComponents(reference, {});
     component.imply("meridiem", Meridiem.AM);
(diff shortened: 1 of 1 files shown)


wanasit:
Sorry for my very late reply. I'm not a native English speaker, but my impression is that "midnight" is a specific time of the day (at 00:00), while "noon" feels like a rough estimate around noon. 

If you confirm, noon should mean be assigned (rather than implied) to 12. I'm ok with accepting this change in English.

--

And, yes the minute and second you probably follow assign/imply setting for the hour.

crazyjncsu:
I consider midnight and noon equal in specificity

Issue #600, opened 2025-06-06
How do I import the ParsingContext?


wickedest:
I am using version v2.8.2 and am trying to implement a `Parser`. However, I cannot figure out how to import the `ParsingContext` type (or the `RegExpMatchArray` type). It does not appear to be exponsed (afaik).

wanasit:
Yes. #601 has been merged.

PR #601, opened 2025-06-07
fix: export type ParsingContext

wickedest:


wickedest:
Fixes #600

wickedest:
I did my best to improve coverage, but I didn't change anything that would cause the coverage to [drop](https://coveralls.io/builds/74042181). It says `ZHHansWeekdayParser.ts` has not been touched in 3 years. There must be some other issue causing the drop.

PR #637, opened 2026-01-17
Fix meridiem merge for casual times and add -5d midnight regression

VergilYotov:
- Original bug: When merging a relative date (-5d) with a time-only expression (00), the date component’s implied meridiem (PM) could override the time parser’s implied AM, shifting midnight to noon. This also
    caused casual phrases like “tonight at 8” to lose their PM context in other locales.  Example -5d 00 was resolving to 5 days ago at 12:00:00 (noon).

  - Fix: Meridiem is now only overridden by the time side when the date side lacks a meaningful meridiem (certain or tagged casual reference). Casual time parsers for es/pt/nl now emit casualReference/* tags so
    their implied PM/AM is respected. 

  - New test: Added a regression case for "-5d 00" to ensure it resolves to Sep 26, 2016 00:00 from an Oct 1, 2016 noon reference.

  Please review the merge heuristic; if you’d like a different “meaningful meridiem” rule (e.g., stricter tags or locale-specific treatment), I can adjust.

### DIFF
--- a/src/calculation/mergingCalculation.ts
+++ b/src/calculation/mergingCalculation.ts
@@ -64,9 +64,14 @@ export function mergeDateTimeComponent(
         dateTimeComponent.assign("timezoneOffset", timeComponent.get("timezoneOffset"));
     }
 
+    const dateHasMeaningfulMeridiem =
+        dateComponent.get("meridiem") != null &&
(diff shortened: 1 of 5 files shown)


wanasit:
Sorry for my slow reply. Thank you for investigating and fixing this.

PR #644, opened 2026-05-05
fix(en): support 'of' connector in weekday postfix modifier (e.g. 'Tuesday of next week')

breissingerj:
## Problem

`ENWeekdayParser` did not recognise the word `of` as a connector between a weekday and its postfix modifier (`this/last/next week`). The postfix group regex only allowed optional whitespace directly before the modifier keyword:

```ts
// before
"(?:\\s*(this|last|past|next)\\s*week)?"
```

As a result, any phrase of the form **`<weekday> of <modifier> week`** silently dropped the modifier and fell back to `getDaysToWeekdayClosest()`. When today is the named weekday (e.g. parsing `"Tuesday of next week"` on a Tuesday), the closest match is 0 days away — returning *today* instead of next week. Additionally, because the full phrase was not consumed as one token, adding a time component (e.g. `"Tuesday of next week after 2pm"`) produced **two separate results** instead of one merged date-time.

## Fix

Add `(?:of\\s*)?` as an optional non-capturing group inside the postfix:

```ts
// after
"(?:\\s*(?:of\\s*)?(this|last|past|next)\\s*week)?"
```

This accepts `"Tuesday of next week"`, `"Friday of last week"`, and `"Wednesday of this week"` while leaving all existing patterns (`"next Tuesday"`, `"Tuesday next week"`, `"Tuesday, next week"`, etc.) completely unchanged.

## Tests

A new `test` block — *"Test - Weekday 'of next/last/this week' connector"* — is added to `test/en/en_weekday.test.ts` with 10 cases covering:

- `of next week` for the same weekday as the ref date (the original edge case)
- `of next week` for earlier and later weekdays
- `of last week` for same and different weekdays
- `of this week` for the same and a forward weekday
- Combined with a time component (`after 2pm`, `at 9am`) — asserts a **single** merged result
- In-sentence context

All 519 tests pass (`npm test`).

### DIFF
--- a/src/locales/en/parsers/ENWeekdayParser.ts
+++ b/src/locales/en/parsers/ENWeekdayParser.ts
@@ -12,7 +12,7 @@ const PATTERN = new RegExp(
         "(?:(this|last|past|next)\\s*)?" +
         `(${matchAnyPattern(WEEKDAY_DICTIONARY)}|weekend|weekday)` +
         "(?:\\s*(?:\\,|\\)|\\）))?" +
-        "(?:\\s*(this|last|past|next)\\s*week)?" +
+        "(?:\\s*(?:of\\s*)?(this|last|past|next)\\s*week)?" +
         "(?=\\W|$)",
     "i"
 );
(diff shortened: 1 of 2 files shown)


wanasit:
Thanks for the improvement!

PR #648, opened 2026-06-18
fix: parse year-before-month-name expressions (e.g. "2024 Aug")

### LINKED ISSUE #639: Bug report : Incorrect parsing of `yyyy MMM` format
# Summary
When parsing "2024 Aug", The date selected is August 2025. `yyyy MMM` is a common format. 

# Steps to reproduce

Sandbox Link : 
https://codesandbox.io/p/devbox/natural-time-string-parsing-with-chrono-node-forked-zrxpj7?workspaceId=ws_34u1DDsW9TLNxiFeaK5zqV

# Reproducer:
```
import "./styles.css";
import * as chrono from "chrono-node";

const dateparse = (dateString) => {
  const parsedDate = chrono.parseDate(dateString);
  if (!parsedDate) return dateString;
  return parsedDate.toLocaleDateString();
};

export default function App() {
  const times = ["2024 Aug", "2024 August", "2023 Sept"];

  return (
    <dl className="App">
      {times.map((time) => (
        <>
          <dt>{time}</dt>
          <dd>= {dateparse(time)}</dd>
          <br />
        </>
      ))}
    </dl>
  );
}
```

# Expected behavior
I expected the parsed result to be date object that corresponds to August 2024.

# Environment
chrono-node version: 2.3.4
Node.js version: v20.14.0
Platform: Windows / macOS / Linux (include version)
Locale / language: e.g. en, en_GB, etc.
Timezone: e.g. Asia/Kolkata

chatman-media:
Closes #639

## Problem

Parsing `"2024 Aug"` returns **August 2023** instead of August 2024 (the `yyyy MMM` order is a common format). Reproduced with a pinned reference date:

```js
chrono.parse("2024 Aug", new Date(2024, 0, 15));
// => 2023-08-20  ❌  (expected 2024-08-01)
```

Two parsers were involved:

1. **`ENMonthNameParser`** only recognized a *trailing* year (`MMM yyyy`), so for `"2024 Aug"` it matched just `"Aug"`, found no year, and fell back to `findYearClosestToRef` — which, from a January reference, picks the *previous* August.
2. **`ENMonthNameLittleEndianParser`** then matched the contiguous `2024` as a bogus day range: `20` + (empty connector) + `24` → "the 20th–24th of August". Because both results spanned the same text, overlap resolution kept this wrong one.

The reverse order (`"Aug 2024"`) already worked, which is why this looked inconsistent.

## Fix

- **`ENMonthNameParser`**: recognize an optional leading 4-digit year, so `"2024 Aug"`, `"2024-Aug"`, `"2024.Aug"`, and `"2012 January"` parse as a single result with a *certain* year. A 4-digit year is required (not 2-digit) to avoid clashing with the day-month format (`"09 Aug"` must stay "9 August"). A guard rejects supplying a year on both sides (`"2012 January 2013"`).
- **`ENMonthNameLittleEndianParser`**: require a real separator between the two day numbers of a range, so a contiguous 4-digit number is no longer split into a spurious `day-to-day` range. Existing ranges are unaffected: `"10 - 22 August"`, `"10 to 22 August"`, `"10-22 August"`, and space-only `"10 22 August"` all still parse as ranges.

## Tests

Added a `Year-Month expression (year before month)` block in `test/en/en_month.test.ts` covering `"2024 Aug"`, `"2024 August"`, `"2024-Aug"`, and `"2012 January"` with pinned reference dates. These **fail** on `master` (`Expected: 2024, Received: 2023`) and **pass** with the fix.

Full suite green: 615 passed.

### DIFF
--- a/src/locales/en/parsers/ENMonthNameLittleEndianParser.ts
+++ b/src/locales/en/parsers/ENMonthNameLittleEndianParser.ts
@@ -12,7 +12,7 @@ const PATTERN = new RegExp(
     `(?:on\\s{0,3})?` +
         `(${ORDINAL_NUMBER_PATTERN})` +
         `(?:` +
-            `\\s{0,3}(?:to|\\-|\\–|until|through|till)?\\s{0,3}` +
+            `(?:\\s{0,3}(?:to|\\-|\\–|until|through|till)\\s{0,3}|\\s{1,3})` +
             `(${ORDINAL_NUMBER_PATTERN})` +
         ")?" +
         `(?:-|/|\\s{0,3}(?:of)?\\s{0,3})` +
--- a/src/locales/en/parsers/ENMonthNameParser.ts
+++ b/src/locales/en/parsers/ENMonthNameParser.ts
@@ -7,6 +7,7 @@ import { AbstractParserWithWordBoundaryChecking } from "../../../common/parsers/
 
 const PATTERN = new RegExp(
     `((?:in)\\s*)?` +
+        `(?:([0-9]{4})\\s*[-.\\/]?\\s*)?` +
         `(${matchAnyPattern(MONTH_DICTIONARY)})` +
         `\\s*` +
         `(?:` +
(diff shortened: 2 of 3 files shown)


wanasit:
Thank you for fixing the issue. Could you please address one comment?

Issue #590, opened 2025-03-31
Concerns & flaws with timezone parsing & handling


LJNeon:
I've noticed that there are quite a few open issues that involve timezones (#385, #403, #532, #546, #552, #564, #582, & possibly more).

Almost all of these would be fixed with three changes:
- Use the IANA timezone database.
- Treat timezones differently based on their source, e.g. parsed from natural language or timezones supplied to the library as an option.
- Detect entirely relative times when parsing natural language and don't apply timezones to them. (And also add a matching boolean property to the output?)

Using the IANA timezone database would allow you to support longer timezone codes like `"America/Los_Angeles"` as well as correctly handle dates in the past when different daylight savings rules were in effect.
Separated timezone logic would resolve #564: CET would be handled with daylight savings if passed as part of the input string and handled as an absolute time if passed as a timezone. This would account for the common usage as well as the expectation of developers.
And not applying the timezone handling to relative times when unnecessary would prevent parsing issues related to relative times and timezones from even occurring in the first place.

After doing some research and experimentation I found the [timezonecomplete](https://www.npmjs.com/package/timezonecomplete) library on npm and it seems to be exactly what would be needed to handle all of these issues. It would also be very nice to be able to outsource timezone logic to a dedicated library due to how complicated and prone to change it is.

I'm willing to open a PR (or several) for this, would just want to confirm how they should behave first.

kimjuan:
```
    // Note: Many sources define CET as a constant UTC+1. In common usage, however,
    // CET usually refers to the time observed in most of Europe, be it standard time or daylight saving time.
```

No one living in CET can code better than @wanasit and authors of this wonderful project, but anyone living with DST will know the above is incorrect.

"CET refers to the time observed" Only In Wikipedia.

If use as a literal in a datetime, CET (or any abbreviation) is always an alias to an constant offset. #564 is absolutely spotted on.

If I watch TV commercial "Live on 25th April 15:00 CET". I will call their hotline,
Either one of the two will happen:

1. _Yes, we mean CET._
2. _Thank you for spotting this typo, we will correct it to CEST asap._

Under no scenario, _Dear, don't you know CET in April auto compute to CEST, like Europe/Berlin_.

That maybe why IANA is invented and what IANA is for.

There are many countries today, past, future, using CET but doesn't observe DST. To use this parser-powered datetime-picker, can imagine every ~March-to-July, their users need to manually offset 1hour from their input, such that it refers to that specific moment in their mind.

[...]

PR #650, opened 2026-06-27
Fix DST resolution depending on the system timezone

### LINKED ISSUE #465: Specific date with hour is parsed to null
I use `chrono-node` version `2.3.9`. I ran some tests with different date formats and found a specific combination of date and hour to not work. I used `parseDate("03-27-2022, 02:00 AM")`.

**Results in null**:
03-27-2022, 02:00 AM

**Working as expected**:
03-27-2022, 01:00 AM
03-27-2022, 03:00 AM
03-27-2022, 04:00 AM
...

03-27-2022, 01:00 PM
03-27-2022, 02:00 PM
03-27-2022, 03:00 PM
...

03-26-2022, 02:00 AM
03-28-2022, 02:00 AM

elinathan:
## Summary

`chrono` resolves some times differently, or drops them entirely, depending on the timezone of the machine running it. The affected times are those that fall in the **system** timezone's DST spring-forward gap.

```js
// On a machine in America/New_York (or any zone with that gap):
chrono.parseDate("2022-03-13T02:30:00-05:00").toISOString();
// => "2022-03-13T08:30:00.000Z"   ❌ off by one hour
// On a UTC machine:
// => "2022-03-13T07:30:00.000Z"   ✅

// Reproduces #465 — returns null on UTC+1 machines, since 02:00 is the CET gap:
chrono.parseDate("03-27-2022, 02:00 AM"); // => null
```

## Cause

`ParsingComponents.date()` and `isValidDate()` build dates with the local `new Date(year, month, …)` constructor and then correct with `getTimezoneOffset()`. Inside the system timezone's DST gap, the offset the constructor used and the offset at the resulting instant disagree, so the instant comes out wrong, and `isValidDate()` (a hard filter in `UnlikelyFormatFilter`) rejects the result, dropping it.

## Fix

Build the instant from `Date.UTC(...)` when a timezone offset is known, and validate the calendar in UTC. The output is now deterministic regardless of host timezone. When no offset is known anywhere, the previous system-local behavior is preserved. For all non-gap inputs the result is unchanged (the existing test suite passes as-is).

## Tests

Adds timezone-parameterized regression tests (UTC / New York / Chicago / Berlin / Tokyo) covering both the resolved instant and calendar validity. Both fail on `master` and pass with this change.

_Testing note:_ Jest can't change the process timezone mid-run. Assigning `process.env.TZ` doesn't invalidate V8's cached zone, so the test only actively asserts on a host whose local timezone has the gap, and is a no-op on the UTC CI runner (consistent with the existing test). If you'd prefer a test that guards this in CI regardless of runner timezone, I can switch it to spawn child processes that run under several real `TZ` values.

Fixes #465. Related to #590.

### DIFF
--- a/src/results.ts
+++ b/src/results.ts
@@ -232,13 +232,26 @@ export class ParsingComponents implements ParsedComponents {
     }
 
     isValidDate(): boolean {
-        const date = this.dateWithoutTimezoneAdjustment();
+        // Check calendar validity in UTC. The local Date constructor would wrongly reject real
+        // times falling in the system timezone's DST gap, making validity depend on where it runs.
+        const date = new Date(
+            Date.UTC(
+                this.get("year"),
+                this.get("month") - 1,
+                this.get("day"),
+                this.get("hour"),
+                this.get("minute"),
+                this.get("second"),
+                this.get("millisecond")
+            )
+        );
+        date.setUTCFullYear(this.get("year"));
 
-        if (date.getFullYear() !== this.get("year")) return false;
-        if (date.getMonth() !== this.get("month") - 1) return false;
-        if (date.getDate() !== this.get("day")) return false;
-        if (this.get("hour") != null && date.getHours() != this.get("hour")) return false;
-        if (this.get("minute") != null && date.getMinutes() != this.get("minute")) return false;
+        if (date.getUTCFullYear() !== this.get("year")) return false;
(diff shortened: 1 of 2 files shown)


wanasit:
Thank you for reporting and fixing this. 

Although we have extensively tested this, I'll publish this in the next minor version (v9.10.0) just in case.

PR #651, opened 2026-06-29
fix: keep casual-time meridiem in de/fr/it/fi when merged with an explicit time

mahirhir:
When a date part and a time part are merged, `mergeDateTimeComponent` decides whether the explicit time's meridiem may override the date part's meridiem. Since #637 it keeps the date part's meridiem only when that meridiem is certain or carries a `casualReference/` tag:

```ts
const dateHasMeaningfulMeridiem =
    dateComponent.get("meridiem") != null &&
    (dateComponent.isCertain("meridiem") ||
        Array.from(dateComponent.tags()).some((t) => t.startsWith("casualReference/")));
```

#637 added that tag to the inline casual-time parsers for `es`, `nl` and `pt`, and the shared `common/casualReferences` helpers already carry it (which covers `en`). The inline casual-time parsers for `de`, `fr`, `it` and `fi` were not updated, so their casual references only imply a meridiem without the tag. When such a reference is merged with a bare time, the time's implied AM wins and the afternoon/evening hint is dropped:

- `de`: "nachmittag um 5" gives 05:00 instead of 17:00
- `fr`: "après-midi à 5" gives 05:00 instead of 17:00
- `it`: "pomeriggio alle 5" gives 05:00 instead of 17:00
- `fi`: "iltapäivällä klo 5" gives 05:00 instead of 17:00

This adds the matching `casualReference/...` tag to each casual branch in those four parsers, the same way #637 did for the other locales, plus a regression test per locale. `en`/`es`/`nl`/`pt` already parse these correctly.

The full suite passes (618 tests). Reverting just the parser changes makes the four new tests fail with `Expected: 17, Received: 5`.

### DIFF
--- a/src/locales/de/parsers/DECasualTimeParser.ts
+++ b/src/locales/de/parsers/DECasualTimeParser.ts
@@ -24,6 +24,7 @@ export default class DECasualTimeParser extends AbstractParserWithWordBoundaryCh
                 component.imply("minute", 0);
                 component.imply("second", 0);
                 component.imply("meridiem", Meridiem.AM);
+                component.addTag("casualReference/morning");
                 break;
 
             case "vormittag":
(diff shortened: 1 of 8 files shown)


wanasit:
Thanks for the fix.

PR #652, opened 2026-06-29
fix(uk): compile forwardDate "within" pattern with the unicode flag

mahirhir:
`UKTimeUnitWithinFormatParser` builds its `forwardDate` branch with `new RegExp(PATTERN, "i")`, dropping the `u` flag.

`PATTERN` ends with `REGEX_PARTS.rightBoundary` (`(?=[^\p{L}\p{N}_]|$)`), and `patternLeftBoundary()` returns `REGEX_PARTS.leftBoundary` (`([^\p{L}\p{N}_]|^)`), so the regex `AbstractParserWithWordBoundaryChecking.pattern()` compiles for this parser contains `\p{L}`/`\p{N}`. Without the `u` flag those are read as identity escapes (`p`, `{`, `L`, `}`, ...), not Unicode letter/number classes, so the word boundary stops rejecting Cyrillic letters.

The non-forward branch right below it already uses `REGEX_PARTS.flags` (`"iu"`), and both branches of the Russian `RUTimeUnitWithinFormatParser` use `REGEX_PARTS.flags`, so the `"i"` here looks like a copy that was left behind.

In practice, with `forwardDate: true` a unit that is only a prefix of a longer word still matches. `годин` (hours) is a prefix of `годинників` (watches), so `купив 5 годинників` ("bought 5 watches") is parsed as `5 годин` ("5 hours").

The change uses `REGEX_PARTS.flags` in the forward branch, the same as the sibling branch and the `ru` parser. I added a regression to `test/uk/uk_time_units_within.test.ts`; `test/uk` and `test/ru` are green (86 tests).

### DIFF
--- a/src/locales/uk/parsers/UKTimeUnitWithinFormatParser.ts
+++ b/src/locales/uk/parsers/UKTimeUnitWithinFormatParser.ts
@@ -12,7 +12,7 @@ export default class UKTimeUnitWithinFormatParser extends AbstractParserWithWord
 
     innerPattern(context: ParsingContext): RegExp {
         return context.option.forwardDate
-            ? new RegExp(PATTERN, "i")
+            ? new RegExp(PATTERN, REGEX_PARTS.flags)
             : new RegExp(`(?:протягом|на протязі|протягом|упродовж|впродовж)\\s*${PATTERN}`, REGEX_PARTS.flags);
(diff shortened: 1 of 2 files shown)


wanasit:
Thanks for the change.

PR #659, opened 2026-09-09
fix(en): reject partial month/year matches in invalid dates

### LINKED ISSUE #513: Why is '50/6/2018' being parsed as Jun 01 2018?
If we have a date which doesn't make sense: for example `50/6/2018`

then it's being parsed as `Jun 01 2018`.

Shouldn't this give `undefined`? Or is there a way to determine the confidence? Or disable this behavior?

ryanchou1994:
Fixes #513.

Parsing `50/6/2018` currently extracts `6/2018` and returns June 1, 2018. Check the original text before a month/year match, following the boundary rule already used by `SlashDateFormatParser`, so this invalid date is skipped. Checking for a preceding digit also handles retries inside a rejected match such as `99/06/2018`.

The regressions cover casual, strict, and GB parsing. They also check valid month/year and complete date inputs, exact text/index boundaries, and continued parsing of a valid date after a rejected one. The change is limited to the English month/year parser and its existing slash-date test file.

Validation:

- Workspace `npm test`: original 146 suites / 680 tests passed; final 146 suites / 686 tests passed on macOS and Linux.
- The focused regressions fail against the original parser (6 failed / 12 passed) and pass with the fix (18 passed) on both platforms.
- CommonJS/ESM builds, TypeScript checking, Prettier, and 24 supplemental assertions through the built CommonJS/ESM exports passed on both platforms. Linux verification ran in a container with networking disabled.
- ESLint remains blocked by the existing configuration: the normal script cannot find an ESLint 9 flat config, and legacy mode references the removed `prettier/@typescript-eslint` export. Both failures also occur on the unchanged base; no dependency or lint configuration changes are included.


ryanchou1994:
Hi @wanasit, friendly ping: the workflows on this PR are still waiting for approval to run. Happy to adjust anything if needed. Thanks!

PR #661, opened 2026-09-19
fix(en): ignore for-the-unit phrases regardless of case

jakezwang:
Make the existing `for the <unit>` exclusion case-insensitive, matching the parser's main pattern. `For the year` and `FOR THE YEAR` currently produce a spurious relative date, including an extra result in uppercase financial statements.

Add mixed-case negative cases and an embedded month/year regression.

Validation: all 146 suites / 684 tests and CommonJS/ESM builds pass on Node 22 and 24; formatting and the pre-commit checks pass. `npm run eslint` remains blocked by the existing missing ESLint 9 flat configuration (#606).


wanasit:
Thank you for the improvement.

